PR #1, opened 2015-08-19
command line completion for gitup command

goranche:
added a simple command line completion script; added code to install.sh, which loads the command line completion script from /etc/bashrc

### DIFF
--- a/Tool/gitup-completion.sh
+++ b/Tool/gitup-completion.sh
@@ -1,10 +1,10 @@
 #!/bin/sh -e
 function _gitup() {
-	local cur opts
-	COMPREPLY=()
-	cur="${COMP_WORDS[COMP_CWORD]}"
-	opts="help open map commit stash"
-	COMPREPLY=($(compgen -W "$opts" -- $cur))
-	return 0
+  local cur opts
+  COMPREPLY=()
+  cur="${COMP_WORDS[COMP_CWORD]}"
+  opts="help open map commit stash"
+  COMPREPLY=($(compgen -W "$opts" -- $cur))
+  return 0
 }
 complete -F _gitup gitup
--- a/Tool/gitup-completion.sh
+++ b/Tool/gitup-completion.sh
@@ -1,0 +1,10 @@
+#!/bin/sh -e
+function _gitup() {
+	local cur opts
+	COMPREPLY=()
+	cur="${COMP_WORDS[COMP_CWORD]}"
+	opts="help open map commit stash"
+	COMPREPLY=($(compgen -W "$opts" -- $cur))
+	return 0
+}
+complete -F _gitup gitup
--- a/Tool/install.sh
+++ b/Tool/install.sh
@@ -7,7 +7,11 @@ INSTALL_DIR=`dirname "$INSTALL_PATH"`
 mkdir -p "$INSTALL_DIR"
 ln -sf "$TOOL_PATH" "$INSTALL_PATH"
 
+# Check if gitup-completion is present in /etc/bashrc ( grep -q ...),
+# otherwise (grep || ...) add our completion loading script to /etc/bashrc
+# more info on here documents: http://tldp.org/LDP/abs/html/here-docs.html
 grep -q gitup-completion /etc/bashrc || cat << EOF >> /etc/bashrc
+# load gitup-completion.sh if it exists; enables command line completion
 gitup="${TOOL_PATH}-completion.sh"; [ -f "\${gitup}" ] && source "\${gitup}"
 EOF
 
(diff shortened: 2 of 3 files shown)


Tool/gitup-completion.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh -e
+function _gitup() {
+	local cur opts
```
swisspol: GitUp uses double-spaces instead of tabs everywhere

goranche: ugh, sorry... :blush: 
will fix

Tool/install.sh
```diff
@@ -7,4 +7,8 @@ INSTALL_DIR=`dirname "$INSTALL_PATH"`
 mkdir -p "$INSTALL_DIR"
 ln -sf "$TOOL_PATH" "$INSTALL_PATH"
```
swisspol: This would need some detailed comments explaining what exactly this does as this is some pretty advanced bash scripting.

swisspol:
Thinking more about this, the general concern I have is that installing a symlink in `/usr/bin` is one thing, and easily revertable too, but patching `/etc/bashrc` is another. At the minimum the user should be prompted to allow this extra operation.

And then, there's this problem: http://apple.stackexchange.com/questions/193368/what-is-the-rootless-feature-in-el-capitan-really

Tool/install.sh
goranche: done... hopefully :smiley:

goranche:
> installing a symlink in /usr/bin is one thing, and easily revertable too, but patching /etc/bashrc is another

ok, not much I can do here :smiley: I personally don't see it that way, I don't really have a problem with adding functionality, especially since there are no side effects even if GitUp is removed from the system.

it'd be better if bash on OS X was set up like on Ubuntu, but unfortunately it isn't :unamused: 

regarding the RootLess feature in El Capitan, I don't really see an issue here, I don't see it changing anything that is being done here

swisspol:
Isn't `/etc/bashrc` not writable in El Capitan?

goranche:
> Isn't /etc/bashrc not writable in El Capitan?

:thought_balloon:  AFAIK no... from the article you linked:

> You can't modify anything in /System, /bin, /sbin, or /usr (except /usr/local).
> ...
> But since normal OS X-style customizations go in /Library (or ~/Library, or /Applications), and unix-style customizations (e.g. Homebrew) go in /usr/local (or sometimes /etc), this shouldn't be a big deal.

as I understand it, nothing changes for "normal" locations, including ```/etc```

I really hope I'm not wrong, because in that case a lot of my other scripts will break as well, including my Xcode server integrations :worried:

fnky:
I don't like the approach of overwriting someones `.bashrc`. Also this wouldn't work if the user uses a shell that does not run `.bashrc`.

Plus there's a lot of developers who doesn't use the `*rc` file for scripts like these, but rather have modularized their dotfiles into plugins/topics and only use the `*rc` file as bootstrap.

goranche:
> I don't like the approach of overwriting bashrc

ok, cool, just close the PR then :unamused: :+1: 

:wink:

fnky:
You're right, should have been less egoistic, but I meant “It's usually not a good idea to ...”

goranche:
don't worry about it, I meant my response as a silly one :blush: :wink: 

I too would love it if Apple had configured bash to use ```/etc/bash_completion.d``` (like Debian), so we could just create a file named ```/etc/bash_completion.d/gitup``` and be done with it... but unfortunately, they didn't :unamused:   and as it is, I don't really see a better way of doing this on OS X

but of course, it's a small feature, doubt too many people would really even notice it :innocent:

fnky:
What could be done is having documentation, [similar to Git](https://git-scm.com/book/en/v1/Git-Basics-Tips-and-Tricks) to inform how users can install auto-completion, so it's more of an opt-in approach rather than forcing it on them.

goranche:
true, that would be an option as well...

:thought_balloon: should the auto-complete script be left in the GitUp.app bundle, or move it out as well?

in that case we would instruct the user to source the script in the bundle, which would allow for upgrades should the gitup command gain options

fnky:
While it is easier to update the app from developer-perspective, what would happen if the app is deleted or moved from location, and possibly other reasons I can't think off my head at the moment.

goranche:
it just will not load (I like to use ```[ -f ... ]``` checks)

fnky:
Then it could a possible solution :-)

swisspol:
> in that case we would instruct the user to source the script in the bundle, which would allow for upgrades should the gitup command gain options

Yes, that's already how it works for the tool: it's just symlink'ed, which makes guarantees it's always in sync with the app. Yes it would break if people relocate the app after installing the tool, but that's not a problem in real-life.

swisspol:
Looks like there's no user friendly and reliable way to do this, so I'm going to close this. Thanks for taking a pass at it though! :smile:

PR #3, opened 2015-08-19
Faster graph rendering for repositories with huge amount of branches.

mandrigin:
We have a huge repo at work (~22000 branches at the moment) and GitUp is really slow on it.

While profiling it turns to be that GitUp makes a lot of comparisons while building an object graph.
instruments2 2015-08-19 14-05-05

I have made a special optimised class which is much faster (also it checks SHA strings, not objects). According to Instruments, it is not the heaviest stack trace anymore (and the app seems to be far more usable on our repo right now).

### DIFF
--- a/GitUpKit/Core/GCOrderedSet.h
+++ b/GitUpKit/Core/GCOrderedSet.h
@@ -1,0 +1,42 @@
+//  Copyright (C) 2015 Pierre-Olivier Latour <info@pol-online.net>
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#import <Foundation/Foundation.h>
+
+@class GCObject;
+
+/**
+ * This class is optimized to be fast when you have a lot of calls to
+ * - addObject: , -containsObject: and -removeObject: methods.
+ */
+@interface GCOrderedSet : NSObject
+
+/**
+ * Accessing this property is CPU-expensive.
+ * It makes a copy of internal storage, filtered by existing objects.
+ * Try to store the value somewhere else and don't access this property if you don't have to.
+ */
+@property(nonatomic, readonly) NSArray* objects;
+
+/**
+ * NOTE: Usually it is unnecessary to add an object, then remove it, then add it again.
+ * But if you will do this, it will appear in the object array
+ * at the SAME PLACE AS IT WAS ADDED.
+ */
+- (void)addObject:(GCObject*)object;
+- (BOOL)containsObject:(GCObject*)object;
+- (void)removeObject:(GCObject*)object;
+
+@end
--- a/GitUpKit/Core/GCOrderedSet.m
+++ b/GitUpKit/Core/GCOrderedSet.m
@@ -1,0 +1,76 @@
+//  Copyright (C) 2015 Pierre-Olivier Latour <info@pol-online.net>
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#if !__has_feature(objc_arc)
+#error This file requires ARC
+#endif
+
+#import "GCPrivate.h"
+
+@implementation GCOrderedSet {
+  NSMutableArray* _objects; // Contains all the objects, even removed ones
+  CFMutableSetRef _actualObjectHashes; // Objects that were added but have not been removed
+  CFMutableSetRef _removedObjectHashes;
+}
+
+- (instancetype)init {
+  if ((self = [super init])) {
+    _objects = [[NSMutableArray alloc] init];
+    _actualObjectHashes = CFSetCreateMutable(kCFAllocatorDefault, 0, NULL);
+    _removedObjectHashes = CFSetCreateMutable(kCFAllocatorDefault, 0, NULL);
+  }
+  return self;
+}
+
+- (void)dealloc {
+  CFRelease(_actualObjectHashes);
+  CFRelease(_removedObjectHashes);
+}
+
+- (void)addObject:(GCObject*)object {
+  if (![self containsObject:object]) {
+    if (CFSetContainsValue(_removedObjectHashes, (__bridge const void*)(object.SHA1))) {
+      CFSetRemoveValue(_removedObjectHashes, (__bridge const void*)(object.SHA1));
+    } else {
+      [_objects addObject:object];
+    }
+    CFSetAddValue(_actualObjectHashes, (__bridge const void*)(object.SHA1));
+  }
+}
+
+- (void)removeObject:(GCObject*)object {
+  if ([self containsObject:object]) {
+    // Removing object from NSMutableArray is expensive,
+    // so we just moving SHA from one set to another.
+    CFSetRemoveValue(_actualObjectHashes, (__bridge const void*)(object.SHA1));
+    CFSetAddValue(_removedObjectHashes, (__bridge const void*)(object.SHA1));
+  }
+}
+
+- (BOOL)containsObject:(GCObject*)object {
+  return CFSetContainsValue(_actualObjectHashes, (__bridge const void*)(object.SHA1));
+}
+
+- (NSArray*)objects {
+  NSMutableArray* result = [[NSMutableArray alloc] initWithCapacity:_objects.count];
+  for (GCObject* object in _objects) {
+    if ([self containsObject:object]) { // Return only objects that were not removed
+      [result addObject:object];
+    }
+  }
+  return result;
+}
+
+@end
--- a/GitUpKit/Core/GCCore.h
+++ b/GitUpKit/Core/GCCore.h
@@ -27,6 +27,7 @@
 #endif
 #import "GCMacros.h"
 #import "GCObject.h"
+#import "GCOrderedSet.h"
 #import "GCReference.h"
 #import "GCSnapshot.h"
 #import "GCReferenceTransform.h"
--- a/GitUpKit/Core/GCOrderedSet-Tests.m
+++ b/GitUpKit/Core/GCOrderedSet-Tests.m
@@ -1,0 +1,87 @@
+//  Copyright (C) 2015 Pierre-Olivier Latour <info@pol-online.net>
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#import "GCTestCase.h"
+#import "GCOrderedSet.h"
+
+@implementation GCMultipleCommitsRepositoryTests(GCOrderedSetTests)
+
+- (void)testOrderedSetAddObject {
+  GCOrderedSet* collection = [[GCOrderedSet alloc] init];
+  [collection addObject:self.commit1];
+  XCTAssertTrue([collection containsObject:self.commit1]);
+  NSArray* commits = collection.objects;
+  XCTAssertEqual(commits.count, 1);
+  XCTAssertEqual(commits[0], self.commit1);
+}
+
+- (void)testOrderedSetAddDuplicateObject {
+  GCOrderedSet* collection = [[GCOrderedSet alloc] init];
+  [collection addObject:self.commit1];
+  [collection addObject:self.commit1];
+  [collection addObject:self.commit2];
+  XCTAssertTrue([collection containsObject:self.commit1]);
+  NSArray* commits = collection.objects;
+  XCTAssertEqual(commits.count, 2);
+  XCTAssertEqual(commits[0], self.commit1);
+  XCTAssertEqual(commits[1], self.commit2);
+}
+
+- (void)testOrderedSetRemoveObject {
+  GCOrderedSet* collection = [[GCOrderedSet alloc] init];
+  [collection addObject:self.commit1];
+  [collection addObject:self.commit2];
+  XCTAssertTrue([collection containsObject:self.commit1]);
+  XCTAssertTrue([collection containsObject:self.commit2]);
+  [collection removeObject:self.commit1];
+  XCTAssertFalse([collection containsObject:self.commit1]);
+  XCTAssertTrue([collection containsObject:self.commit2]);
+  NSArray* commits = collection.objects;
+  XCTAssertEqual(commits.count, 1);
+  XCTAssertEqual(commits[0], self.commit2);
+}
+
+- (void)testOrderedSetReAddObject {
+  GCOrderedSet* collection = [[GCOrderedSet alloc] init];
+  [collection addObject:self.commit1];
+  [collection addObject:self.commit2];
+  XCTAssertTrue([collection containsObject:self.commit1]);
+  XCTAssertTrue([collection containsObject:self.commit2]);
+  [collection removeObject:self.commit1];
+  XCTAssertFalse([collection containsObject:self.commit1]);
+  XCTAssertTrue([collection containsObject:self.commit2]);
+  [collection addObject:self.commit1];
+  XCTAssertTrue([collection containsObject:self.commit1]);
+  XCTAssertTrue([collection containsObject:self.commit2]);
+  // NOTE: this collection preserves the original place of commit if re-added
+  NSArray* commits = collection.objects;
+  XCTAssertEqual(commits.count, 2);
+  XCTAssertEqual(commits[0], self.commit1);
+  XCTAssertEqual(commits[1], self.commit2);
+}
+
+- (void)testOrderedSetObjectsOrdering {
+  GCOrderedSet* collection = [[GCOrderedSet alloc] init];
+  [collection addObject:self.commit1];
+  [collection addObject:self.commit2];
+  [collection addObject:self.commit3];
+  NSArray* commits = collection.objects;
+  XCTAssertEqual(commits.count, 3);
+  XCTAssertEqual(commits[0], self.commit1);
+  XCTAssertEqual(commits[1], self.commit2);
+  XCTAssertEqual(commits[2], self.commit3);
+}
+
+@end
--- a/GitUpKit/GitUpKit.xcodeproj/project.pbxproj
+++ b/GitUpKit/GitUpKit.xcodeproj/project.pbxproj
@@ -7,6 +7,11 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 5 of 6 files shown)


swisspol:
GitUp is not designed for very large repos - see http://forums.gitup.co/t/gitup-limitations-known-issues/29.

This change will make things a bit faster, but the reason it's an NSMutableArray right now and not hashed-based container is because ordering of tips must be preserved.

mandrigin:
I'm still profiling the code, but this change reduced the startup time on my computer from ~1 minute to ~10 seconds which is already a good thing, IMO. I would love to be able to use it at work, it is a really cool client :)

Ok, so the hash table approach per se doesn't work here.
What if we will use a hash table or set only to check if commit is already there? 
E.g.:
````
if (![tipsDict objectForKey:commit.SHA1]) {
    [tipsArray addObject:commit];
    tipsDict[commit.SHA1] = @(YES);
}
````
Can there be any case when the same commit will be in this array twice?

**Comment by mandrigin** [comment]
Of course I will encapsulate the code in a separate class that will both preserve commit order and have a faster way to check if commit has already been added.

**Comment by swisspol** [comment]
I unfortunately don't have bandwidth to help understand the code. It's been written to be pretty clear regarding how it works and you can always trace what is happening using the debugger :smile:

**Comment by mandrigin** [comment]
Sure, no problem :)

**Comment by bryankeller** [comment]
Perhaps a separate data structure could be used to keep track of ordering, while still seeing the speed benefits from using the NSMutableDictionary?

**Comment by mandrigin** [comment]
Yes, I'm currently working on it.

/ Igor M

**Comment by mandrigin** [comment]
Ok, tests are passing now, but I have no idea why they were failing in commit `ed9b145`. I did nothing except forcing Travis to re-test branch. So I think it is pretty much ready IMO. I'll just rename the last commit.

**Comment by swisspol** [comment]
FYI I updated the contribution requirements at https://github.com/git-up/GitUp#contributing.

**Comment by mandrigin** [comment]
:+1: Will fix my pull request then.

**Comment by mandrigin** [comment]
Done

**Comment by swisspol** [comment]
So what's the performance like with this change? GitUp will log exactly how long graph generation and graph rendering take.

Note that like I was saying GitUp is not designed to handle very large repos, so a number of things are going to be inefficient, not just this. Improving the speed from say 1mn to 10s is interesting but doesn't solve the actual problem: for it to be usable it needs to be below 1s.

**Comment by swisspol** [comment]
Also there are still a number of areas that don't match the GitUp style e.g. `foo *bar` instead of `foo* bar`, missing braces for `if()` statements, ...

**Comment by mandrigin** [comment]
Test set: the repository of ~237000 commits. I run it on my MacBook Pro (2015).

The performance before this change:
`00:01:26.214 [VERBOSE  ]> Graph regenerated for "<redacted>" in 76.468 seconds`

The performance after this change:
`00:00:23.048 [VERBOSE  ]> Graph regenerated for "<redacted>" in 3.623 seconds`

It is still not that comfortable, but looks like a good step forward.

Yes, there are a few places that can be improved, but I wanted to take an iterative approach and try to optimise one thing at a time, until it gets usable on the repo of our size. Maybe it is very ambitious, but at least I have a very good test repo :)

I will fix codestyle issues and also I noticed one warning I have introduced, will fix that as well.

**Comment by mandrigin** [comment]
Done. I tried to imitate the code style you have in an app, but please, tell me if anything left...

File: Core/GCObject.h [hunk 1]
```diff
@@ -15,6 +15,8 @@
 
 #import <Foundation/Foundation.h>
 
+@class GCRepository;
```
swisspol: This should be a separate commit.

Core/GCTestCase.h
```diff
@@ -79,6 +79,9 @@
 @interface GCMultipleCommitsRepositoryTests : GCMultipleCommitsRepositoryTestCase
 @end
 
+@interface GCUniqueCommitCollectionTests : GCMultipleCommitsRepositoryTestCase
```
swisspol: You don't need a new top-level test class. You only need a category on `GCMultipleCommitsRepositoryTestCase`.

Core/GCUniqueCommitCollection-Tests.m
```diff
@@ -0,0 +1,87 @@
...
+#import "GCTestCase.h"
+#import "GCUniqueCommitCollection.h"
+
+@implementation GCUniqueCommitCollectionTests (GCUniqueCommitCollectionTests)
+
+- (void)testAdd {
+  GCUniqueCommitCollection* collection = [GCUniqueCommitCollection new];
+  [collection addCommit:self.commit1];
+  XCTAssertTrue([collection hasCommit:self.commit1]);
+  NSArray *commits = collection.commits;
```
swisspol: * position

Core/GCUniqueCommitCollection-Tests.m
```diff
@@ -0,0 +1,87 @@
...
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#import "GCTestCase.h"
+#import "GCUniqueCommitCollection.h"
+
+@implementation GCUniqueCommitCollectionTests (GCUniqueCommitCollectionTests)
+
+- (void)testAdd {
+  GCUniqueCommitCollection* collection = [GCUniqueCommitCollection new];
```
swisspol: Pattern in GitUp is to use -alloc + -init, not +new.

Core/GCUniqueCommitCollection.h
```diff
@@ -0,0 +1,43 @@
...
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#import <Foundation/Foundation.h>
+
+@class GCCommit;
+
+
```
swisspol: Extra blank line

Core/GCUniqueCommitCollection.h
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * This class represents a collection of commits with preserved order.
+ * It is also optimized to be fast when you have a lot of calls to
+ * - addCommit: , -hasCommit: and -removeCommit: methods.
+ */
+@interface GCUniqueCommitCollection : NSObject
```
swisspol: It doesn't seem this class has anything to do with commits: it should be called `GCOrderedSet` as it is fundamentally an ordered set unless I missed something.

Core/GCUniqueCommitCollection.h
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * NB! Accessing this property is CPU-expensive.
+ * It makes a copy of internal storage, filtered by existing commits.
+ * Try to store the value somewhere else and don't access this property if you don't have to.
+ */
+@property (nonatomic, readonly) NSArray* commits;
```
swisspol: Extra space between `property` and `(`

Core/GCUniqueCommitCollection.m
```diff
@@ -0,0 +1,71 @@
...
+
+@implementation GCUniqueCommitCollection {
+  NSMutableArray* _commits; // Contains all the commits, even removed ones
+  NSMutableSet* _actualCommitHashes; // Commits that were added but have not been removed
+  NSMutableSet* _removedCommitHashes;
+}
+
+- (instancetype)init {
+  if ((self = [super init])) {
+    _commits = [NSMutableArray array];
```
swisspol: Use alloc / init

Core/GCUniqueCommitCollection.h
swisspol: As a matter of fact it doesn't need to manipulate `GCCommit`s, only `id`s.

Core/GCUniqueCommitCollection.m
```diff
@@ -0,0 +1,71 @@
...
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#import "GCUniqueCommitCollection.h"
```
swisspol: This missing the ARC #ifdef like in other .m files.

Interface/GIGraph.m
```diff
@@ -313,19 +306,20 @@ - (void)_generateGraph {
   }
   
   // Make sure we have some tips left
-  if (tips.count == 0) {
+  if (tips.commits.count == 0) {
     goto cleanup;
   }
   
   // Re-sort all tips in descending chronological order (this ensures virtual tips will be on the rightside of the tips descending from the same commits)
+  NSMutableArray *tipsArray = tips.commits.mutableCopy;
```
swisspol: `*` position

Interface/GIGraph.m
```diff
@@ -18,6 +18,7 @@
 #endif
 
 #import "GIPrivate.h"
+#import "GCUniqueCommitCollection.h"
```
swisspol: Remove this and add header to `GCCore.h` instead.

Interface/GIGraph.m
```diff
@@ -313,19 +306,20 @@ - (void)_generateGraph {
   }
   
   // Make sure we have some tips left
-  if (tips.count == 0) {
+  if (tips.commits.count == 0) {
```
swisspol: You said above that `-commits` is expensive: why generate an entire array just to get the count?

Interface/GIGraph.m
swisspol: This is not efficient: just grab the array and create a sorted copy only if sorting is needed.

Core/GCUniqueCommitCollection.m
```diff
@@ -0,0 +1,71 @@
...
+}
+
+- (BOOL)hasCommit:(GCCommit*)commit {
+  return [_actualCommitHashes containsObject:commit.SHA1];
+}
+
+- (NSArray*)commits {
+  NSMutableArray* result = [NSMutableArray array];
+  for (GCCommit* commit in _commits) {
+    if ([self hasCommit:commit]) { // Return only commits that were not removed
```
swisspol: No need to call -hasCommit, access `_actualCommitHashes` directly.

Core/GCUniqueCommitCollection.m
```diff
@@ -0,0 +1,71 @@
...
+  [_actualCommitHashes removeObject:commit.SHA1];
+  [_removedCommitHashes addObject:commit.SHA1];
+}
+
+- (BOOL)hasCommit:(GCCommit*)commit {
+  return [_actualCommitHashes containsObject:commit.SHA1];
+}
+
+- (NSArray*)commits {
+  NSMutableArray* result = [NSMutableArray array];
```
swisspol: This should have upfront the capacity of the total number of commits in the collection.

Core/GCUniqueCommitCollection.m
```diff
@@ -0,0 +1,71 @@
...
+  if ((self = [super init])) {
+    _commits = [NSMutableArray array];
+    _actualCommitHashes = [NSMutableSet new];
+    _removedCommitHashes = [NSMutableSet new];
+  }
+  return self;
+}
+
+- (void)addCommit:(GCCommit *)commit {
+  if (![self hasCommit:commit]) {
```
swisspol: No need to call -hasCommit, access _actualCommitHashes directly.

Core/GCUniqueCommitCollection.m
```diff
@@ -0,0 +1,71 @@
...
+      [_removedCommitHashes removeObject:commit.SHA1];
+    } else {
+      [_commits addObject:commit];
+    }
+    [_actualCommitHashes addObject:commit.SHA1];
+  }
+}
+
+- (void)removeCommit:(GCCommit*)commit {
+  if (![self hasCommit:commit]) {
```
swisspol: Use consistent patterns: either `if (!...) return` or `if (...) { }` but don't mix and match as you are doing with `-addCommit:` and `-removeCommit:`.

swisspol:
I had a first pass look and pointed out a few things. You need to keep crafting this whole thing until it is as concise and elegant as possible - keep iterating! :smile: 

For instance, I would use CFMutableSets with no retain / release callbacks (since the commits are already in the array anyway and therefore retained there) and low-overhead hash/equal callbacks accessing directly the underlying `git_object`. Retain/release becomes expensive when dealing with lots of objects. Same for using ARC in this file. If it's that performance sensitive, then MRC will be much better (critical code paths in GitUp are all MRC).

In any case what makes you think the number of commits in your repo is the problem here? This code only cares about tips i.e. references. What's the output of `git show-ref --head | wc -l` in your repo?

Core/GCUniqueCommitCollection.m
swisspol: The `-addCommit:` pattern is preferred in GitUp.

Core/GCTestCase.h
mandrigin: That was my first approach. But then I found out that I have to call methods like `testCommitsRepositoryAddCommit` and so forth. Because `testAdd` makes no sense on the class named `GCMultipleCommitsRepositoryTestCase`, it is misleading.

Then I decided to make this class and I like how it actually looks in XCode testing section.
gctestcase m 2015-08-21 08-10-45

swisspol: Yes it looks better in the Xcode UI, but that's not following the GitUp existing convention of how tests are structured, which is required :smile:

mandrigin: Okay, will fix it. Consistency matters ;)

Core/GCUniqueCommitCollection.h
mandrigin: Current implementation is a quite generic one, I agree.
The difference between ordered set is that when you re-add a commit it preserves the old position. It doesn't really matter in our case (we do not re-add commits, if memory serves), but it might cause some issues in future. I don't know if it is worth making too generic... But it might be a good point, I will think about it.

It also uses SHA1 method,  so it probably should accept `GCObject*` then, not `id`.

Interface/GIGraph.m
mandrigin: ¯\_(ツ)_/¯ that is my fault.

mandrigin:
> In any case what makes you think the number of commits in your repo is the problem here?
Yes, we have a lot of tips in our repo as well. Anyway it looks like a good test set for this product :)

> What's the output of `git show-ref --head | wc -l` in your repo?
~22000

Core/GCUniqueCommitCollection.m
mandrigin: What's the reason for it?
For me `[self hasCommit:commit]` is a shorter and more maintainable thing than using `[_actualCommitHashes containsObject:commit.SHA1]`. 
Like if we want to change the "backend" store for this class and replace `NSMutableSet` if will be less places to make the same changes.
(Maybe I'm overthinking it).

mandrigin:
I've fixed most of the things you pointed.
This thread might be also a good reference to the code style you have...

> For instance, I would use CFMutableSets with no retain / release callbacks (since the commits are already in the array anyway and therefore retained there) and low-overhead hash/equal callbacks accessing directly the underlying git_object. Retain/release becomes expensive when dealing with lots of objects. Same for using ARC in this file. If it's that performance sensitive, then MRC will be much better (critical code paths in GitUp are all MRC).

Currently I see only a small overhead from this class in profiler. 
These changes are really reasonable but is it worth to implement them right now?

GitUpKit/Core/GCOrderedSet.m
```diff
@@ -0,0 +1,73 @@
...
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#if !__has_feature(objc_arc)
+#error This file requires ARC
+#endif
+
+#import "GCOrderedSet.h"
+
```
swisspol: Replace this by `#import GCPrivate.h` like other implementation files.

GitUpKit/Core/GCOrderedSet.h
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * This class is optimized to be fast when you have a lot of calls to
+ * - addObject: , -containsObject: and -removeObject: methods.
+ */
+@interface GCOrderedSet : NSObject
+
+/**
+ * NB! Accessing this property is CPU-expensive.
+ * It makes a copy of internal storage, filtered by existing commits.
```
swisspol: Docs still say "commit"

GitUpKit/Core/GCOrderedSet.h
```diff
@@ -0,0 +1,42 @@
...
+ * Try to store the value somewhere else and don't access this property if you don't have to.
+ */
+@property(nonatomic, readonly) NSArray* objects;
+
+/**
+ * NOTE: Usually it is unnecessary to add a commit, then remove it, then add it again.
+ * But if you will do this, it will appear in commits array 
+ * in the SAME PLACE AS IT WAS ADDED.
+ */
+- (void)addObject:(GCObject*)commit;
```
swisspol: Argument should be "object" instead of "commit"

swisspol:
> These changes are really reasonable but is it worth to implement them right now?

If this is a critical code path, which it is, then I would say yes, might as well go 100%, instead of stopping at 90% :smile:

In any case this PR is certainly a step in the right direction, but still need some polish before it can land. Keep reviewing that diff over and over again until you're 100% confident this is it.

Also note that the problem here is not the number of commits but the number of branches in your repo (22,000), which is plain huge. In my experience however, repos that have such an abnormally high number of branches are also going to have hundreds of thousands of commits if not millions, so other parts of GitUp become really slow.

BTW the commit title should be more precise since this change optimizes only drawing graphs with lots of refs, not huge repos in general.

mandrigin:
I did some measurements, and found out that I can get a somewhat visible performance improvements only by using `CFMutableSet` class. Neither moving to MRC, nor accessing `object.private` didn't help much. (Maybe I did something wrong there though, I can't say I'm an expert in MRC).

> In my experience however, repos that have such an abnormally high number of branches are also going to have hundreds of thousands of commits if not millions, so other parts of GitUp become really slow.

I did some more profiling and the next important bottlenecks are:

- History scanning is not very slow (~5 seconds). I didn't look at it precisely yet, though I think there are some places where `GCOrderedSet` might help as well.

- Checking changes in the work directory. This case is more tricky, and it is not *that* bad (3~4 seconds), but still quite annoying. I don't have a good solution yet, but I'll try to understand what can be done here as well.

swisspol:
The problem with ARC when you have code dealing with *lots* of objects in tight loops or similar is that since it has to be extremely conservative to avoid crashes, it retains / releases everything. So you end up with hundreds of thousands or retain/release calls which take locks and slow down everything.

A repo can be huge along 3 axis (or any combination of course):

1. Number of commits (> 100,000)
2. Number of refs (> 1,000)
3. Number of files in the trees (> 10,000)

Each of them will affect GitUp differently, requires appropriate performance analysis, and dedicated fixes. My understanding is that you're hitting (2).

The important thing is that GitUp has been *deliberately* architectured to handle very efficiently repos below these limits. In the real world, the *vast vast majority* of repos are an order of magnitude smaller than these limits.

The direct consequences is that properly addressing these performance issues (and going from say 10s to 2s is still not enough as a reactive app needs to be <1s on representative hardware) is most likely going to require rewriting entire parts of GitUpKit, which may be a lot of work and is certainly very high-risk in terms of potential regressions.

It's interesting work no matter what, so by all means, feel free to continue, but I thought that it's important you understand what you're getting into :smile:

mandrigin:
I do agree, making app fast on huge amount of data is hard. 

There are 3 types of performance fixes that can be introduced for an app:
1) Real performance improvement w/o re-designing the app (by introducing faster data types, by getting rid of unnecessary re-checks, etc).
2) Real performance improvement by redesigning the app architecture.
3) Perceived performance improvements (progress-bars, animations, making the UI responsive even though the app is busy, etc). There is a book written on this topic:  [Designing and Engineering Time](http://www.amazon.com/Designing-Engineering-Time-Psychology-Perception/dp/0321509188)

I will try to focus on (1) first.

> A repo can be huge along 3 axis (and any combination of course):

I guess our repo hit all 3 of this... Just refs had the worst performance penalty.

mandrigin:
Ok, I did change the commit message, rebased it on the latest version of master. Is there anything left to be done?

swisspol:
It's looking pretty good now. Just some minor things left.

Did you make to add the new `.m` file to the iOS GitUpKit target as well?

Please agree to GitUp CLA at http://forums.gitup.co/t/gitup-contributor-license-agreement/323.

mandrigin:
> Did you make sure to add the new .m file to the iOS GitUpKit target as well?

Yes, I did ;)
screenshot

> Please agree to GitUp CLA at http://forums.gitup.co/t/gitup-contributor-license-agreement/323
Done

swisspol:
Your screenshot is for the OSX target though :)

mandrigin:
Yes, but I just double-checked an iOS version as well :)

swisspol:
All right, great work, thanks!

mandrigin:
Awesome!

swisspol:
The header was missing but I fixed it and added a test in Travis CI to catch this going forward: 9d06064277e77fcba4a1269721748b94637e056c.

mandrigin:
oops...

PR #4, opened 2015-08-19
Simplify search bar placeholder

salutis:
Changes search bar placeholder from "Search Repository..." to simple "Search" to match standard OS X apps like Contacts or Notes.


swisspol:
This is not going to work correctly per https://github.com/git-up/GitUp/blob/master/Application/Document.m#L517. See other locations where the cell string is set.

swisspol:
To be clear, I don't mind this change, but extensive testing needs to be done covering every possible state of the search field and making sure there are no visual regression.

salutis:
Ah, I see. This is way too clever (in a bad way) use of the placeholder. Closing for now then.

PR #7, opened 2015-08-25
Faster git-clone due to depth=1 (not for developers)

denji:
* Immediately pulling all sub-modules


swisspol:
I don't see why this would be necessary: GitUp is actually a very small repo, so cloning time is not an issue.

Besides, telling people to do a shallow clone without explaining what is happening is setting them up for failure. They will end up with half a repo and most Git users have no clue what a shallow clone is (and therefore how to use them).

PR #8, opened 2015-08-25
Refactor GitUpKit Base classes into separate static library

JrGoodle:
The GitUpKit Core and Extension sources are compiled into a static library named libGitUpKitBase and the XLFacility sources are compiled into a static library named libXLFacility. This allows for more modular dependency management in other targets (Tests, GitUpKit Framework). It also makes the base GitUpKit functionality available for use on iOS in the future, since it only depends on Foundation (my primary motivation for these changes was mostly curiosity about what components could be reused on iOS).

There shouldn't be any functional changes to the code; the changes are contained to the Xcode build system.

### DIFF
--- a/GitUp/GitUp.xcodeproj/project.pbxproj
+++ b/GitUp/GitUp.xcodeproj/project.pbxproj
@@ -33,6 +33,20 @@
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
+		C6D9F6621B8B081500695231 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = E28A16DA1B87021600218332 /* GitUpKit.xcodeproj */;
+			proxyType = 2;
+			remoteGlobalIDString = C6D9F5B01B8B05D000695231;
+			remoteInfo = GitUpKitBase;
+		};
+		C6D9F6641B8B081500695231 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = E28A16DA1B87021600218332 /* GitUpKit.xcodeproj */;
+			proxyType = 2;
+			remoteGlobalIDString = C6D9F62F1B8B069200695231;
+			remoteInfo = XLFacility;
+		};
 		E21DCAF01B2538C5006424E8 /* PBXContainerItemProxy */ = {
 			isa = PBXContainerItemProxy;
 			containerPortal = E2C338A319F8562F00063D95 /* Project object */;
@@ -191,7 +205,9 @@
 			isa = PBXGroup;
 			children = (
 				E28A16E01B87021600218332 /* GitUpKit.framework */,
+				C6D9F6631B8B081500695231 /* libGitUpKitBase.a */,
 				E28A16E21B87021600218332 /* GitUpTests.xctest */,
+				C6D9F6651B8B081500695231 /* libXLFacility.a */,
 			);
 			name = Products;
 			sourceTree = "<group>";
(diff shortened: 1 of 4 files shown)


swisspol:
There is no reason to change the products of the Xcode project at this time: a single framework is much simpler and less maintenance.

JrGoodle:
OK. Since the sources were factored into static libraries, the primary product was still just a framework (i.e. no changes were necessary in the GitUp project or example projects). I was thinking it was simpler in the sense that the sources only needed to be compiled into one target, and that target was linked into others; rather than including the sources in multiple targets.

Do you have a preferred workflow for pull requests in terms of logging an issue on the GitUp Forums first or something like that? I don't want to annoy you with too many unwanted PR's.

swisspol:
> I was thinking it was simpler in the sense that the sources only needed to be compiled into one target, and that target was linked into others; rather than including the sources in multiple targets.

But now you have all these extra targets and libraries and whatnot. It's actually more complex than simply having a framework target and an associated test target, the later not even being exposed outside the project.

> Do you have a preferred workflow for pull requests in terms of logging an issue on the GitUp Forums first or something like that?

PRs that just change stuff for no real reason are not going to make it into mainstream :smile: The code is clean and organized already. Bug fixes and well coded enhancements that benefit the majority of users are welcome. For non-trivial stuff, it doesn't hurt to start a discussion on forums.gitup.co first.

JrGoodle:
> PRs that just change stuff for no real reason are not going to make it into mainstream :smile:
This change wasn't exactly for "no real reason", I probably didn't explain my intent very well. I was curious about trying to leverage GitUpKit on iOS. However, since the framework contains both the base and UI layers, there's a dependency on AppKit so it can't be used there. Rather than creating another framework and having to include all the base sources again, I moved them into their own static library so it could be used in the current **GitUpKit.framework**, and possibly reused in a **GitUpKitMobile.framework** or something for iOS.

> The code is clean and organized already.
The code is quite clean and organized; very nice work there :smiley_cat:. It made the creation of a static library for the base classes quite simple, since there was already a clean separation between the base and UI layers in the sources.

swisspol:
As far as the GitUp project is concerned, which is only what matters, since we talking about a PR into upstream, this change is for no real reason: it's just moving files around for no real benefit whatsoever but increased complexity of the project.

In your fork, you can certainly test and refactor things as much as you want, but there is no reason for the results of these tests to make it into upstream. If you had a completely working version of GitUp on iOS, it would be a different story.

JrGoodle:
That's understandable. Not that I'm anywhere near this point, but would a working example project be enough? The prospect of a working version of **GitUp** on iOS is a much more daunting challenge.

swisspol:
You should be able to add an iOS framework target to GitUpKit and have Core + Extensions + Interface (minus the view classes - see `__GI_HAS_APPKIT__` in `GIInterface.h`) in it. Then you can have an iOS app example in `Examples/`. That said I'm not sure there's any point in doing this as it's not like you can do much on iOS with that reduced size framework.

JrGoodle:
Thanks, I had gotten about that far (and came across `__GI_HAS_APPKIT__`). I realized that the use of `NSTask` presents a problem for iOS, with sandboxing and all.

I understand the use case on iOS is much more limited than on OS X. However, I could imagine novel use cases for at least rewriting commits on mobile devices.

swisspol:
NSTask is only used to run Git hooks, which make no sense on iOS anyway for a variety of reasons.

JrGoodle:
Thanks. I made it to the point where I need to compile libgit2 for iOS. I'll take a look into that another day...

PR #13, opened 2015-09-01
Animated transition in/out of Quick View

jtbandes:
As Quick View feels a lot like Quick Look, I thought it would be nice to have some smooth transitions. (It would be nice for the title bar to fade at the same time, but I didn't do that in this commit.)

- Use 2 snapshot layers to animate the transition in/out of Quick View
- Rename `Document`'s `_containerView` to `_mapContainerView`, to make its purpose more evident
- Link QuartzCore.framework

### DIFF
--- a/GitUp/Application/Document.h
+++ b/GitUp/Application/Document.h
@@ -39,7 +39,7 @@ typedef NS_ENUM(NSUInteger, WindowModeID) {
 @property(nonatomic, weak) IBOutlet NSButton* helpOpenButton;
 
 @property(nonatomic, weak) IBOutlet NSTabView* mainTabView;
-@property(nonatomic, weak) IBOutlet NSView* containerView;
+@property(nonatomic, weak) IBOutlet NSView* mapContainerView;
 
 @property(nonatomic, strong) IBOutlet NSView* titleView;
 @property(nonatomic, weak) IBOutlet NSTextField* titleTextField;
--- a/GitUp/Application/Document.h
+++ b/GitUp/Application/Document.h
@@ -29,6 +29,7 @@ typedef NS_ENUM(NSUInteger, WindowModeID) {
 
 @interface Document : NSDocument <NSUserInterfaceValidations>
 @property(nonatomic, strong) IBOutlet NSWindow* mainWindow;
+@property(nonatomic, strong) IBOutlet NSView* contentView;
 
 @property(nonatomic, strong) IBOutlet NSToolbar* toolbar;
 
--- a/GitUp/Application/Document.m
+++ b/GitUp/Application/Document.m
@@ -45,6 +45,10 @@
 #define kRestorableStateKey_WindowMode @"windowMode"
 
 #define kSideViewAnimationDuration 0.15  // seconds
+#define kQuickViewAnimationDuration 0.25  // seconds
+
+#define kAnimationKey_ID @"animationID"
+#define kAnimationID_QuickViewSnapshot @"snapshotAnimation"
 
 #define kMaxAncestorCommits 1000
 
@@ -105,6 +109,8 @@ static inline WindowModeID _WindowModeIDFromString(NSString* mode) {
   BOOL _unifiedToolbar;
   NSNumberFormatter* _numberFormatter;
   NSDateFormatter* _dateFormatter;
+  CALayer* _fixedSnapshotLayer;
+  CALayer* _animatingSnapshotLayer;
   NSMutableArray* _quickViewCommits;
   GCHistoryWalker* _quickViewAncestors;
   GCHistoryWalker* _quickViewDescendants;
@@ -278,7 +284,13 @@ static void _CheckTimerCallBack(CFRunLoopTimerRef timer, void* info) {
   if (_unifiedToolbar) {
     _mainWindow.titleVisibility = NSWindowTitleHidden;
   }
-  _mapContainerView.wantsLayer = YES;
+  _contentView.wantsLayer = YES;
+  
+  // Text fields must be drawn on an opaque background to avoid subpixel antialiasing issues during animation.
+  for (NSTextField* field in @[_infoTextField1, _infoTextField2, _progressTextField]) {
+    field.drawsBackground = YES;
+    field.backgroundColor = _mainWindow.backgroundColor;
+  }
   
   _mapViewController = [[GIMapViewController alloc] initWithRepository:_repository];
   _mapViewController.delegate = self;
@@ -880,11 +892,11 @@ static NSString* _StringFromRepositoryState(GCRepositoryState state) {
     XLOG_DEBUG_UNREACHABLE();
   }
   if (showHelp) {
-    NSRect contentBounds = [_mainWindow.contentView bounds];
+    NSRect contentBounds = _contentView.bounds;
     _helpView.hidden = NO;
     _mainTabView.frame = NSMakeRect(contentBounds.origin.x, contentBounds.origin.y, contentBounds.size.width, contentBounds.size.height - _helpView.frame.size.height);
   } else if (!_helpView.hidden) {
-    _mainTabView.frame = [_mainWindow.contentView bounds];
+    _mainTabView.frame = _contentView.bounds;
     _helpView.hidden = YES;
   }
 }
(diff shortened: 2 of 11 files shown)


GitUp/Application/Document.m
```diff
@@ -880,11 +886,11 @@ - (void)_showHelpWithIdentifier:(NSString*)identifier {
     XLOG_DEBUG_UNREACHABLE();
   }
   if (showHelp) {
-    NSRect contentBounds = [_mainWindow.contentView bounds];
```
swisspol: I had problems with `NSRect rect = view.bounds` that doesn't compile in Xcode 6 or 7, I don't remember, so you have to do do `[view bounds]` instead.

GitUp/Application/Document.h
```diff
@@ -39,7 +40,7 @@ typedef NS_ENUM(NSUInteger, WindowModeID) {
 @property(nonatomic, weak) IBOutlet NSButton* helpOpenButton;
 
 @property(nonatomic, weak) IBOutlet NSTabView* mainTabView;
-@property(nonatomic, weak) IBOutlet NSView* containerView;
+@property(nonatomic, weak) IBOutlet NSView* mapContainerView;
```
swisspol: Can you put this change into a separate prior commit?

GitUp/Application/Document.m
```diff
@@ -936,7 +942,99 @@ - (void)_enterQuickViewWithHistoryCommit:(GCHistoryCommit*)commit commitList:(NS
...
+  }
+  
+  [_contentView.layer addSublayer:_fixedSnapshotLayer];
+  [_contentView.layer addSublayer:_animatingSnapshotLayer];
+  
+  _fixedSnapshotLayer.frame = _contentView.layer.bounds;
+  
+  CALayer* fromLayer;
+  if (animationInFlight) {
+    fromLayer = _animatingSnapshotLayer.presentationLayer ?: _animatingSnapshotLayer;
```
swisspol: Please don't use the shortcut version of the ternary operator as I don't use it elsewhere in the code base.

GitUp/Application/Document.m
```diff
@@ -936,7 +942,99 @@ - (void)_enterQuickViewWithHistoryCommit:(GCHistoryCommit*)commit commitList:(NS
...
+  [group setValue:kAnimationID_QuickViewSnapshot forKey:kAnimationKey_ID];
+  
+  [CATransaction setAnimationDuration:kQuickViewAnimationDuration];
+  [CATransaction setAnimationTimingFunction:[CAMediaTimingFunction functionWithName:kCAMediaTimingFunctionDefault]];
+  [_animatingSnapshotLayer addAnimation:group forKey:kAnimationID_QuickViewSnapshot];
+  
+  [CATransaction commit];
+}
+
+- (void)animationDidStop:(CAAnimation *)anim finished:(BOOL)flag {
```
swisspol: Space before `*`

GitUp/Application/Document.m
```diff
@@ -936,7 +942,99 @@ - (void)_enterQuickViewWithHistoryCommit:(GCHistoryCommit*)commit commitList:(NS
...
+- (void)_configureAnimatingSnapshotLayerForCommit:(GCHistoryCommit*)commit {
+  if (commit) {
+    NSPoint commitPoint = [_contentView convertPoint:[_mapViewController positionForCommit:commit] fromView:_mapViewController.view];
+    if (NSPointInRect(commitPoint, _contentView.bounds)) {
+      _animatingSnapshotLayer.bounds = _contentView.layer.bounds;
+      _animatingSnapshotLayer.position = [_contentView convertPointToLayer:commitPoint];
+    } else {
+      _animatingSnapshotLayer.frame = _contentView.layer.bounds;
+    }
+    _animatingSnapshotLayer.transform = CATransform3DMakeScale(0.0001, 0.0001, 1);
```
swisspol: This needs a comment explaning the `0.001`.

GitUpKit/Utilities/GIAppKit.h
```diff
@@ -36,6 +36,7 @@ extern NSString* const GICommitMessageViewUserDefaultKey_EnableSpellChecking;
 
 @interface NSView (GIAppKit)
 - (void)replaceWithView:(NSView*)view;  // Preserves frame and autoresizing mask
+- (NSImage*)snapshotImage;
```
swisspol: The name is a bit ambiguous: is snapshot a verb here or a noun? If the latter, this should be a property instead.

Maybe `-(NSImage*)takeSnapshot`?

GitUpKit/Views/GIMapViewController.m
```diff
@@ -204,6 +204,12 @@ - (GINode*)nodeForCommit:(GCCommit*)commit {
   return historyCommit ? [_graphView.graph nodeForCommit:historyCommit] : nil;
 }
 
+- (NSPoint)positionForCommit:(GCHistoryCommit*)historyCommit {
+  GINode* node = [_graphView.graph nodeForCommit:historyCommit];
+  XLOG_CHECK(node);
```
swisspol: This must be `XLOG_DCHECK()` so that it only asserts in "Debug" builds.

swisspol:
Looks pretty good, that's a well written patch! :smile: 

Some high-level comments:

- Something like this which is a "real" contribution cannot land until I figure out http://forums.gitup.co/t/understanding-the-gitup-open-source-license/317
- If this transition becomes animated, then the window toolbar contents, which also changes, should be animated as well: otherwise it just looks weird
- GitUp is compatible with 10.8 and 10.9: does it work there as well?

**The most important factor here is that GitUp does not animate its UI on purpose: it's all about speed and not having superfluous / distracting UI niceties. Although you've made Quick View look "nicer", it's definitely slower and therefore less responsive. I'm not fundamentally against such "enhancements" but they would need to be thought-through and most likely be disabled by default: maybe a "Enable superfluous animations" preference?**

GitUpKit/Views/GIMapViewController.h
```diff
@@ -37,6 +37,7 @@
 @property(nonatomic) BOOL forceShowAllTips;
 - (BOOL)selectCommit:(GCCommit*)commit;  // Also scrolls if needed to ensure commit is visible - Returns YES if commit was selected
 - (GINode*)nodeForCommit:(GCCommit*)commit;
+- (NSPoint)positionForCommit:(GCHistoryCommit*)historyCommit;
```
swisspol: This should take a `GCCommit` for consistency with the other APIs and I think the name should reflect better what's the coordinate system e.g. `-positionInViewForCommit:`.

GitUp/Application/Document.m
jtbandes: I tried 6 and 7 and it seems to work.

`window.contentView.bounds` didn't work because `contentView` was `id` until the 10.11 SDK (when it becomes `__kindof NSView*`).

jtbandes:
> the window toolbar contents, which also changes, should be animated as well

Agreed; I didn't do it in this commit but I can add another.

> GitUp is compatible with 10.8 and 10.9: does it work there as well?

Do you have recommendations for how to test this? The machines I have access to daily are 10.10 and 10.11 only.

As for a preference, I thought about that too. Something akin to OS X's

image

Happy to add that in another commit.

swisspol:
When I (rarely) need to double-check that something works on older OSes, I use Parallels with 10.8 and 10.9 virtual machines. You can easily get the OS images if you are an Apple registered developer.

swisspol:
For the preference, I think it would be OFF by default and since the animations should be OFF by default, that means its wording should be like "Enable visual effects".

jtbandes:
Took a bit of effort to make them run in a VM, but I got there eventually :smiley: 

After a "fun" debugging session I determined that the GIDualSplitView implementation of `splitView:resizeSubviewsWithOldSize:` is causing some perpetual layout-dirtying on 10.8 which makes it hang while trying to commit a transaction. Since there's no comment it's hard to tell what this method is trying to achieve anyway. Any thoughts for an alternate implementation? Can this just be implemented with a holding priority? (It's hard to tell exactly what it's trying to do with no comment :grimacing:)

swisspol:
I thought about it afterwards, but I'm pretty sure you can also use the free Virtual Box to test with older OS X versions.

I wouldn't worry about 10.8 if things don't work there, just disable the code path entirely so restore the exact original behavior:

    if (kCFCoreFoundationVersionNumber >= kCFCoreFoundationVersionNumber10_9) {
      // do animation
    }

Regarding `splitView:resizeSubviewsWithOldSize:`, I don't recall exactly but it was working around some AppKit bug or issue - and yes, I should have written a comment!

jtbandes:
Found it :wink:  http://stackoverflow.com/a/30494691

jtbandes:
Animating toolbar contents will be hard (lack of public API to get at the toolbar's view; and the fact that implicit animations don't work when `hidden` is involved). I think it's best left as a future improvement.

swisspol:
Good find WRT to the splitview :wink: and glad you were able to find a workaround for 10.8!

swisspol:
> I think it's best left as a future improvement.

Well, that's an unfortunate tricky situation then. You have a very nice patch for sure, but the fact remains that from a UX perspective it's half-done because of the toolbar. I'd like to see this landing in some form, but it can't right now.

Can you think of any other way to smooth the toolbar change of appearance?

jtbandes:
Turns out it wasn't as hard as I envisioned :smiley:

swisspol:
Nice! What about 10.8 too? :smile:

swisspol:
The pref commit needs an updated title and it should be before the 2 animation commits.

jtbandes:
Yep, tested in 10.8 and 10.9. Update the commits now...

swisspol:
I just tested the final version and it looks pretty sweet! Great work!

One minor thing: the status text rendering changes briefly during the animation:

Normal:

image

During the animation:

image

Looks like the text labels become their own layer and subpixel font smoothing gets disabled or something.

swisspol:
Also when you have a minute, please accept the new GitUp CLA at http://forums.gitup.co/t/gitup-contributor-license-agreement/323.

swisspol:
This is all ready to land now. The very last thing is that text rendering I was pointing out? Do you have an idea what is causing this and if it can be avoided? If not, I don't have a problem landing it as-is.

jtbandes:
Yes, I modified the commit to include a fix: https://github.com/jtbandes/GitUp/commit/fd4681ce07f0281934171b9922725d9e7e3a0006#diff-9640dac2231f3c7e87bbda41062cc757R289

I noticed there's a similar issue with the "Show" button, but it looks like it's already rendering incorrectly (no subpixel AA) in the release version of GitUp. For some reason, during the animation it *does* have subpixel AA. I don't think there's a similar fix since there is no `drawsBackground` property on that button. Also, Interface Builder lists the Bevel button style under "Discouraged Styles", so I'm inclined to think it may be somewhat unsupported...

swisspol:
All right, it's merged, congrats!

jtbandes:
:confetti_ball:

salutis:
Looks awesome, thanks @jtbandes!

PR #15, opened 2015-09-20
Include both commit messages in squash

sephlietz:
Squash With Parent populates the commit message with the squashed
commit's message and its parent's commit message. This is similar to the
way 'git rebase --interactive' works.

Resolves: http://forums.gitup.co/t/squash-with-parent-should-include-both-commit-messages/315


swisspol:
Sorry, but this does not meet the GitUp quality requirements to be merged.

It doesn't seem this PR was tested as GitUp doesn't handle `#`-prefixed lines in commit message, `_CleanedUpCommitMessage()` is intended to be used on commit messages, not concatenations of them, it's not using NSLocalizedString()` like the rest of the code...

sephlietz:
Thanks for the feedback.

PR #16, opened 2015-09-22
Include both commit messages in squash

sephlietz:
Squash With Parent populates the commit message with the squashed
commit's message and its parent's commit message. This is similar to
the way 'git rebase --interactive' works.

Resolves: http://forums.gitup.co/t/squash-with-parent-should-include-both-commit-messages/315

screen shot 2015-09-21 at 11 24 56 pm

### DIFF
--- a/GitUpKit/Views/GIMapViewController+Operations.m
+++ b/GitUpKit/Views/GIMapViewController+Operations.m
@@ -225,7 +225,9 @@ static inline NSString* _CleanedUpCommitMessage(NSString* message) {
 
 - (void)squashCommitWithParent:(GCHistoryCommit*)commit {
   if ([self checkCleanRepositoryForOperationOnCommit:commit]) {
-    [self _promptForCommitMessage:_CleanedUpCommitMessage(commit.message)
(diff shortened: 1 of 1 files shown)


GitUpKit/Views/GIMapViewController+Operations.m
```diff
@@ -225,7 +225,9 @@ - (void)swapCommitWithChild:(GCHistoryCommit*)commit {
 
 - (void)squashCommitWithParent:(GCHistoryCommit*)commit {
   if ([self checkCleanRepositoryForOperationOnCommit:commit]) {
-    [self _promptForCommitMessage:_CleanedUpCommitMessage(commit.message)
+    GCHistoryCommit* parentCommit = commit.parents.firstObject;
```
swisspol: What happens if the commit has no parent?

sephlietz: In the GitUp UI, the "Squash With Parent" action is disabled when a commit does not have a parent. Outside of GitUp usage this of course result in an error but I believe there are other places that assume `commit.parents` will be non-null.

swisspol:
All good then, thx!

PR #17, opened 2015-10-01
Display both author and committer dates in Quick View

00Davo:
Quick View should include all salient information on the current commit, and previously the author date was completely hidden everywhere in GitUp; this adds separate fields for both kinds of date and displays them simultaneously.

Additionally, as the previously-included date field is no longer required, it has been removed and the space used to show the complete SHA-1 hash of the commit rather than the shortened version.

### DIFF
--- a/GitUpKit/Core/GCCommit.m
+++ b/GitUpKit/Core/GCCommit.m
@@ -63,10 +63,8 @@ static inline NSString* _ConvertMessage(GCCommit* commit, const char* message, s
   return _ConvertMessage(self, summary, strlen(summary), git_commit_message_encoding((git_commit*)_private));
 }
 
-// Reimplementation of git_commit_time()
 - (NSDate*)date {
-  const git_signature* signature = git_commit_committer((git_commit*)_private);
-  return [NSDate dateWithTimeIntervalSince1970:signature->when.time];
+  return self.committerDate;
 }
 
 // Reimplementation of git_commit_time_offset()
@@ -85,6 +83,11 @@ static inline NSString* _ConvertMessage(GCCommit* commit, const char* message, s
   return [NSString stringWithUTF8String:signature->email];
 }
 
+- (NSDate*)authorDate {
+  const git_signature* signature = git_commit_author((git_commit*)_private);
+  return [NSDate dateWithTimeIntervalSince1970:signature->when.time];
+}
+
 - (NSString*)committerName {
   const git_signature* signature = git_commit_committer((git_commit*)_private);
   return [NSString stringWithUTF8String:signature->name];
@@ -93,6 +96,12 @@ static inline NSString* _ConvertMessage(GCCommit* commit, const char* message, s
 - (NSString*)committerEmail {
   const git_signature* signature = git_commit_committer((git_commit*)_private);
   return [NSString stringWithUTF8String:signature->email];
+}
+
+// Reimplementation of git_commit_time()
+- (NSDate*)committerDate {
+  const git_signature* signature = git_commit_committer((git_commit*)_private);
+  return [NSDate dateWithTimeIntervalSince1970:signature->when.time];
 }
 
 - (NSString*)treeSHA1 {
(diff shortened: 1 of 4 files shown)


GitUpKit/Core/GCCommit.m
```diff
@@ -63,10 +63,8 @@ - (NSString*)summary {
   return _ConvertMessage(self, summary, strlen(summary), git_commit_message_encoding((git_commit*)_private));
 }
 
-// Reimplementation of git_commit_time()
 - (NSDate*)date {
-  const git_signature* signature = git_commit_committer((git_commit*)_private);
-  return [NSDate dateWithTimeIntervalSince1970:signature->when.time];
+  return [self committerDate];
```
swisspol: This should be using the dot notation.

GitUpKit/Views/GIQuickViewController.m
```diff
@@ -158,9 +160,11 @@ - (void)setCommit:(GCHistoryCommit*)commit {
       [_diffFilesViewController setDeltas:_diff.deltas usingConflicts:nil];
     } else {
       _sha1TextField.stringValue = @"";
-      _dateTextField.stringValue = @"";
+      _committerDateTextField.stringValue = @"";
```
swisspol: This line is duplicated below, so it should be removed.

GitUpKit/Views/en.lproj/GIQuickViewController.xib
```diff
@@ -49,15 +50,15 @@
                                                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                                                         <subviews>
                                                             <view id="Bpc-ZA-3gm" customClass="GIFlippedView">
-                                                                <rect key="frame" x="0.0" y="0.0" width="300" height="100"/>
+                                                                <rect key="frame" x="0.0" y="0.0" width="300" height="150"/>
```
swisspol: A 50px increase seems a lot, can you provide a screenshot of the updated info area of the XIB?

swisspol:
When editing the XIB, you copy pasted the text fields, right? It has to be done this way to preserve all their settings, which can be tricky.

swisspol:
This PR looks quite good overall BTW, nice work :+1:

00Davo:
Yep, I copied and pasted the text fields and then adjusted the coordinates accordingly. I've pushed fix-upped versions of both commits to address the two style issues.

The XIB looks like this in the builder:

screen shot 2015-10-01 at 3 24 53 pm

And like this when the app's actually running:

screen shot 2015-10-01 at 3 27 01 pm

swisspol:
Ah I see, 50px makes sense as you added 2 fields.

Is selection and copy still working on these fields? I think it was before.

Finally, please reduce the spacing between author and associated date to the standard text field spacing, but leave the spacing between author date and commiter name unchanged. This will make things look more grouped.

00Davo:
Selection and copy isn't working, but it wasn't on the original date field either. It looks like the message field does support selections so I'll see if I can find the relevant properties.

How big's the standard text field spacing? I checked through all the XIB files and there doesn't seem to be anywhere else that has text fields stacked.

swisspol:
Don't worry about selection, if it wasn't enabled already, it's because it must have triggered some rendering bug.

You just drag the text field upwards and it should snap when at the right distance from the above one.

00Davo:
Awesome. I've just pushed the changed layout, which now looks like this:

screen shot 2015-10-01 at 4 16 03 pm

It's definitely much nicer. :cat:

swisspol:
Awesome, thanks!

swisspol:
Actually one last check: did you verify resizing the left pane and that everything was working correctly? The autoresizing flags must be set correctly on the date text fields, and the ones on the sha1 field most likely needs updating.

00Davo:
Um, when the left pane is resized the info stays exactly where it is, rather than resizing with it. Ends up looking like this:

screen shot 2015-10-01 at 4 27 41 pm

This is kind of odd. However, it's also how the app was working beforehand - I just confirmed by building `master` that the exact same behaviour occurs. It might have already regressed?

swisspol:
Yes, it looks like it was never implemented, so no worries, there's no regression here.

PR #18, opened 2015-10-01
Don't block main thread in quick commit view on large commits

nhelke:
The graph view rendering is super fast but does not show any of the commit meta data. This means I often find myself switching through a range of commits in the quick view to find the exact commit I am looking for. The diff computation for large commits locks up the interface. This is a very quick hack I have thrown together. It is not ideal for several reasons (see below) but I thought I would submit it any way to at least start a discussion.

1. Repurposing the emptyLabel requires code duplication and is semantically wrong
2. Firing and forgetting about the work means that the user interface no longer locks 👍 but potentially could consume resources boundlessly if user keeps switching between to large commits without ever letting them finish. Ideally a context could be passed to `diffCommit:withCommit:filePattern:options:maxInterHunkLines:maxContextLines:error:` which would check it in the tightest loop and break out if the context is cancelled (when the quick view moves on to another commit).

### DIFF
--- a/GitUpKit/Views/GIQuickViewController.m
+++ b/GitUpKit/Views/GIQuickViewController.m
@@ -145,19 +145,28 @@ static NSString* _CleanUpCommitMessage(NSString* message) {
       [committer endEditing];
       _committerTextField.attributedStringValue = committer;
       
-      NSError* error;
-      _diff = [self.repository diffCommit:_commit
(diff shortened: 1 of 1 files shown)


swisspol:
Please create a topic at http://forums.gitup.co/c/contributions as that's where discussions on potential PRs should go.

PR #19, opened 2015-10-04
Allow switching to Map/Commit/Stashes directly from Quick View

00Davo:
Cmd+1/2/3 from Quick View previously did absolutely nothing - it'd be more convenient to be able to jump to the other main views instead, and this has been requested in the forum: http://forums.gitup.co/t/324

It might also be worth permitting Cmd+1/2/3 from the "quick diff" view you get with I in the Map. (I'm not sure it has a name?)

### DIFF
--- a/GitUp/Application/Document.m
+++ b/GitUp/Application/Document.m
@@ -1739,7 +1739,7 @@ static NSString* _StringFromRepositoryState(GCRepositoryState state) {
   }
   
   if (item.action == @selector(switchMode:)) {
-    if ([_windowMode isEqualToString:kWindowModeString_Map_QuickView] || [_windowMode isEqualToString:kWindowModeString_Map_Diff]
+    if ([_windowMode isEqualToString:kWindowModeString_Map_Diff]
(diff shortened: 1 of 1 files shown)


swisspol:
How did you test this?

00Davo:
Pretty much just by building and verifying that Cmd+1/2/3 work as expected in Quick View - I don't think there are automated tests for the UI at present?

It does also enable the View menu's Map/Commit/Stashes options while in Quick View, but this is probably desired behaviour anyway.

swisspol:
It's just that all you did was enabling the menu items, apparently ignoring any potential assumptions anywhere else in the code that Quick View would not be active when switching back to the Map view. I wrote this months ago so I don't remember but it's perfectly possible that such assumptions exists and would result in bugs and / or regressions.

For this change to be approved, an audit needs to be designed and performed across the code base for how the "window mode" is used, and making sure nothing would break.

00Davo:
Ah, fair enough. I'm certainly not especially familiar with how the codebase operates overall.

You *can* switch from Quick View to Map View already, so the code should already be able to handle that particular case with no drama, but switching directly from Quick to Stashes or Commit might in some cases have a problem, yeah. Hmm.

swisspol:
Closing stalled pull request.

PR #22, opened 2015-11-23
Improved branch titles

shpakovski:
Improves readability and usability of branch titles
improvement

### DIFF
--- a/GitUpKit/Interface/GIGraphView.m
+++ b/GitUpKit/Interface/GIGraphView.m
@@ -47,8 +47,8 @@
 #endif
 
 #define kTitleSpacing 200
-#define kTitleOffsetX 17
-#define kTitleOffsetY 20
+#define kTitleOffsetX 7
+#define kTitleOffsetY 7
 
 #define kLabelOffsetX 18
 #define kLabelOffsetY 10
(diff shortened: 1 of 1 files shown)


swisspol:
I tried your patch and it looks pretty neat overall, nice work :smile:  Here are some comments:

1) From a UI perspective, it certainly looks quite better than before. I would say however that the text is not as readable as before for a few reasons:
- The text is more rotated and it's a bit harder to read (a few people were complaining about the current rotated text saying it wasn't easy to read but now it's more rotated)
- It seems the font size is a bit smaller and on non-retina displays (I'm using a 27 inch Apple display right now), that makes the non-bold text *very* thin.
- The colored text is a bit too light: it should be a bit more opaque.

image

2) The upstream information for local branches is gone: `{branch} ◄ {upstream}`. This means there's nowhere in the UX now to know what's the upstream for a given branch. That's not *critical* information, but it's useful information and one would expect any Git UI to display it.

If you can't show it in your new UI, then at least modify the contextual menu code around `item = [[NSMenuItem alloc] initWithTitle:NSLocalizedString(@"Set Upstream to", nil) action:NULL keyEquivalent:@""];` so that if an upstream is set, there's a checkmark next to the corresponding branch. This submenu only lists remote branches, and it's possible, although quite rare, to set up another local branch as the upstream. So that menu might have to be changed to:
```
(remote branches)
-separator-
(local branches - excluding the current branch of course) 
```

3) There's a performance regression with the new drawing code. With the old one, `-drawRect:` takes about 2-4ms to render the window at around 1000x700 on my fancy MacBook Pro. With the new code, it takes around 15ms to 45ms, so about 10X slower. I suspected that when simply live resizing the window on its width as it didn't seem as smooth in the new version compared to the old one.

Performance is critical in GitUp so I'm very careful when introducing changes to make sure that they are as optimal as possible, and if they make things slower, that the benefits are worth it.

Instruments shows this:

**New Version**
image

**Old Version**
image

I actually never realized that `CTFontCreateUIFontForLanguage()` was a significant cost so I just did 90ea3d3247f9111b9fe3f45cebdece09926a8b2c to cache the result and it's quite better:

image

4) This 90ea3d3247f9111b9fe3f45cebdece09926a8b2c commit conflicts with your PR. It should be very easy for you to fix the conflict and preserve my change though.

shpakovski:
Thanks!

>The text is more rotated and it's a bit harder to read (a few people were complaining about the current rotated text saying it wasn't easy to read but now it's more rotated).

45˚ makes the underline pixel-perfect and gives more space for titles reducing the risk of label overlapping. As of +5˚, I think it does not make that much difference for people who do not like rotation anyway 🙄

>It seems the font size is a bit smaller and on non-retina displays (I'm using a 27 inch Apple display right now), that makes the non-bold text very thin.

13 pt is too big and most branches with long titles having upstream are cut in our project. The thin part of label is not as meaningful as bold which looks good even on non-Retina displays. So it depends on what information is more important. I find the upstream less useful, and if we draw it in the same bold font, the sense will be lost.

>The colored text is a bit too light: it should be a bit more opaque.

Agree, I made it darker 👌

>The upstream information for local branches is gone: `{branch} ◄ {upstream}`.

No-no, it is not. The format was changed to `{bold branch}\n{thin upstream}`. The image you attached has an example: the top left label. I like the idea with improved menu though, but I believe it is a good candidate for another pull request 🙃

>There's a performance regression with the new drawing code.

Yay, I am caching fonts as well now. Thank you very much for the heads up.

>It should be very easy for you to fix the conflict and preserve my change.

Done. Please check https://github.com/git-up/GitUp/pull/23 Thanks!

PR #23, opened 2015-11-24
New branch titles

shpakovski:
The new format for branch titles which shows more information using branch colors.

### DIFF
--- a/GitUpKit/Interface/GIGraphView.m
+++ b/GitUpKit/Interface/GIGraphView.m
@@ -25,6 +25,7 @@
 
 #define __DEBUG_DRAWING__ 0
 #define __DEBUG_BOXES__ 0
+#define __DEBUG_TITLE_CORNERS__ 0
 #define __DEBUG_MAIN_LINE__ 0
 #define __DEBUG_DESCENDANTS__ 0
 #define __DEBUG_ANCESTORS__ 0
@@ -47,8 +48,8 @@
 #endif
 
 #define kTitleSpacing 200
-#define kTitleOffsetX 17
-#define kTitleOffsetY 20
+#define kTitleOffsetX 7
+#define kTitleOffsetY 7
 
 #define kLabelOffsetX 18
 #define kLabelOffsetY 10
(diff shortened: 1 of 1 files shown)


GitUpKit/Interface/GIGraphView.m
```diff
@@ -860,112 +860,170 @@ static void _DrawLine(GILine* line, CGContextRef context, CGFloat offset, CGFloa
...
+  }
+  static CTFontRef boldFont = NULL;
+  if (boldFont == NULL) {
+    boldFont = CTFontCreateUIFontForLanguage(kCTFontUIFontEmphasizedSystem, 12.0, CFSTR("en-US"));
+  }
+  static NSParagraphStyle* style = nil;
+  if (style == nil) {
+    NSMutableParagraphStyle* mutableStyle = [[NSMutableParagraphStyle alloc] init];
+    mutableStyle.lineHeightMultiple = 0.8;
+    style = [mutableStyle copy];
```
swisspol: There's no need to force the use of a `NSParagraphStyle`. Just keep the `NSMutableParagraphStyle` around.

GitUpKit/Interface/GIGraphView.m
```diff
@@ -860,112 +860,170 @@ static void _DrawLine(GILine* line, CGContextRef context, CGFloat offset, CGFloa
...
+  }
   
-  CFAttributedStringRef string = CFAttributedStringCreate(kCFAllocatorDefault, (CFStringRef)label, (CFDictionaryRef)attributes);
-  CTFramesetterRef framesetter = CTFramesetterCreateWithAttributedString(string);
-  CGSize size = CTFramesetterSuggestFrameSizeWithConstraints(framesetter, CFRangeMake(0, CFAttributedStringGetLength(string)), NULL, CGSizeMake(CGFLOAT_MAX, CGFLOAT_MAX), NULL);
+  // Create a light string to show above the HEAD
+  NSDictionary* multilineTitleAttributes = @{ NSFontAttributeName: (id)titleFont, NSForegroundColorAttributeName: [color shadowWithLevel:0.25], NSParagraphStyleAttributeName: style };
+  NSMutableAttributedString* multilineAttributedTitle = [[NSMutableAttributedString alloc] initWithString:multilineTitle attributes:multilineTitleAttributes];
+  
+  // Change font to bold on ranges collected before
```
swisspol: I didn't notice that before, but you don't need to create a `NSString` then keep around all the attributes, then create a `NSMutableAttributedString` recombining everything. Keep it simple: just do:
```
NSMutableAttributedString* multilineTitle = [[NSMutableAttributedString alloc] init];
(or if you need some base attributes for the entire string)
NSMutableAttributedString* multilineTitle = [[NSMutableAttributedString alloc] initWithString:@"" attributes:...];
[multilineTitle beginEditing];

(loop)
NSInteger length = multilineTitle.length;
[multilineTitle replaceCharactersInRange:NSMakeRange(length, 0) withString:...];
[multilineTitle setAttributes:... range:NSMakeRange(length, multilineTitle.length)];

[beginEditing endEditing];
// draw
[multilineTitle release];
```

swisspol:
*Please don't recreate a new pull request, as we lose the conversation. You can always force push to your branch to update/replace the pull request.*

> No-no, it is not. The format was changed to `{bold branch}\n{thin upstream}`.

Ah, this is confusing. It looks like 2 branches are at the same tip. It needs to be made clear there is a relation between these 2 things.

I tried adding back a few symbols and this is what works best IMO: `@"⬅︎ %@/%@\n"` so please use that.

> 13 pt is too big and most branches with long titles having upstream are cut in our project.

These are 2 separate issues:
- GitUp has been around for a long time and *nobody* has ever complained about 13 pt being too big.
- There will always be clipping of long branch labels. The only way to fix this is not to reduce the font size, but to better compute the required vertical height at the top of the graph to not clip any labels.

In any case, I tried increasing back to size 13 and it's much better. Branch names are pretty important, you shouldn't have to squint to read them. So please change that back.

Finally, you have a bug in your labels: you have local branch names, remote branch names and upstream branches (which can be local or remote but most likely remote). Here the branch tip has `origin/delete_me` and `delete_me` branches at it, and the upstream of `delete_me` is set to `origin/master`, but labels are wrong:

image

Make sure you understand well how the current code gets the names to draw the labels and reproduce *exactly* that.

It's really close though now :smile:

GitUpKit/Interface/GIGraphView.m
shpakovski: Whoa, awesome!

GitUpKit/Interface/GIGraphView.m
shpakovski: Agree, changed.

shpakovski:
>You can always force push to your branch to update/replace the pull request.
My bad, sorry for inconvenience. Now I see, this is *my* repository, so force-push is kind of allowed 😊

>This is what works best IMO: @"⬅︎ %@/%@\n" so please use that.
I tried multiple symbols in the end of the first line as well, but could not find the ideal. Let's use arrow, cool.

>In any case, I tried increasing back to size 13 and it's much better.
Done. I believe labels become too massive next to branch lines, but this effect may be useful by design.

>Finally, you have a bug in your labels…
Thanks! Investigating into this…

swisspol:
Let me know when the label bug is fixed and then I'll do a final review pass.

shpakovski:
Done. Please verify.

shpakovski:
Thanks for comments! I fixed those things and did `push --force`. Please review.

swisspol:
> Sorry for insisting, but I find the font size 12.0 so much nicer than 13.0 😂 Any chance to reconsider or maybe use 13.0 only on non-Retina?

Sorry, 13.0 it is, I don't want to change this part of the UI at this time.

GitUpKit/Interface/GIGraphView.m
```diff
@@ -860,112 +860,164 @@ static void _DrawLine(GILine* line, CGContextRef context, CGFloat offset, CGFloa
...
     CTFrameGetLineOrigins(frame, CFRangeMake(i, 1), &origin);
-    CGContextSetTextPosition(context, textRect.origin.x + origin.x, textRect.origin.y + origin.y);
-    if (size.width <= kMaxBranchTitleWidth) {
+    
+    origin.x += textRect.origin.x + origin.y;
+    origin.y += textRect.origin.y;
+    
+    CGFloat ascent, descent, leading;
+    CGFloat lineWidth = CTLineGetTypographicBounds(line, &ascent, &descent, &leading);
+    CGFloat lineHeight = ascent;
```
swisspol: The line heigh is not `ascent`, it's `ascent` + `descent`. You can remove `lineHeight` and use `ascent` instead.

GitUpKit/Interface/GIGraphView.m
```diff
@@ -860,112 +860,164 @@ static void _DrawLine(GILine* line, CGContextRef context, CGFloat offset, CGFloa
...
+    
+    // Draw separator in case of new line which is guaranteed by the building algorithm
+    CFRange stringRange = CTLineGetStringRange(line);
+    if (stringRange.length == 1) {
```
swisspol: What is this drawing exactly? I don't understand what `(stringRange.length == 1)` is trying to detect?

swisspol:
Actually, I just realized that if you force push, it deletes all the comments in the PR review :disappointed: . So instead of force pushing, just do additional commits. When the PR complete, just flatten everything into a single commit and I'll merge it.

GitUpKit/Interface/GIGraphView.m
```diff
@@ -860,112 +860,164 @@ static void _DrawLine(GILine* line, CGContextRef context, CGFloat offset, CGFloa
...
+      CGRect underlineRect = CGRectMake(origin.x - 1.0, origin.y - 1.0, lastLineWidth + 5.0, lineHeight - 3.0);
```
swisspol: What are the `5.0` and `3.0` for? Do they depend on the font size?

GitUpKit/Interface/GIGraphView.m
```diff
@@ -860,112 +860,164 @@ static void _DrawLine(GILine* line, CGContextRef context, CGFloat offset, CGFloa
...
+    NSRange tagNameRange = NSMakeRange(multilineTitle.length - tagName.length - 2, tagName.length); // -2 to exclude char ']' plus '\n'
+    [multilineTitle addAttribute:NSFontAttributeName value:(id)boldFont range:tagNameRange];
+    [multilineTitle addAttribute:NSForegroundColorAttributeName value:darkColor range:tagNameRange];
   }
-  if (!label.length) {
-    [label release];
-    return;  // This should only happen if we have a detached HEAD with no other references pointing to the commit
+  
+  if (branch.tags.count > 0) {
+    _AppendAttributedString((CFMutableAttributedStringRef)multilineTitle, @"\n", nil);
```
swisspol: Why do you need to add a newline at the end of the string?

GitUpKit/Interface/GIGraphView.m
```diff
@@ -860,112 +860,164 @@ static void _DrawLine(GILine* line, CGContextRef context, CGFloat offset, CGFloa
...
-    } else {
-      if (label.length) {
-        [label appendString:@"\n"];
-      }
-      [label appendString:@"["];
-    }
-    [label appendString:tag.name];
-    ++index;
+  
+  if (branch.remoteBranches.count > 0) {
```
swisspol: This should be `(branch.remoteBranches.count && branch.tags.count)` so that you only insert a newline if there are branches and tags. Then get rid of `_AppendAttributedString((CFMutableAttributedStringRef)multilineTitle, @"\n", nil);` further down.

GitUpKit/Interface/GIGraphView.m
```diff
@@ -860,112 +860,164 @@ static void _DrawLine(GILine* line, CGContextRef context, CGFloat offset, CGFloa
...
   
   // Prepare context
-  
   CGContextSaveGState(context);
   CGContextTranslateCTM(context, x, y);
-  CGContextRotateCTM(context, 40.0 / 180.0 * M_PI);
-  
-  // Draw label
-  
-#if __DEBUG_BOXES__
```
swisspol: Please restore this but fix the code so it does the same thing as before.

GitUpKit/Interface/GIGraphView.m
```diff
@@ -860,112 +860,164 @@ static void _DrawLine(GILine* line, CGContextRef context, CGFloat offset, CGFloa
...
+      
+      NSString* localBranchName = localBranch.name;
+      NSRange localBranchNameRange = NSMakeRange(multilineTitle.length, localBranchName.length);
+      _AppendAttributedString((CFMutableAttributedStringRef)multilineTitle, [NSString stringWithFormat:@"%@\n", localBranchName], multilineTitleAttributes);
+      [multilineTitle addAttribute:NSFontAttributeName value:(id)boldFont range:localBranchNameRange];
+      [multilineTitle addAttribute:NSForegroundColorAttributeName value:darkColor range:localBranchNameRange];
+      
+      // Upstream may be local or remote, so we look for slash to make the branch name more emphasized
+      NSRange upstreamBranchNameRange = NSMakeRange(multilineTitle.length + 2, upstreamName.length); // +2 to exclude "⬅ "
+      NSRange slashRange = [upstreamName rangeOfString:@"/"];
```
swisspol: Use the API here, don't search for `/`. The fact the API uses `-rangeOfString:` internally is an implementation detail. What matters is that we don't have code in the app that duplicates functionality of the API. It's increasing the risk of bugs and pretty much guarantees that improving the API later on or changing its implementation will get things out of sync.

GitUpKit/Interface/GIGraphView.m
```diff
@@ -860,112 +860,164 @@ static void _DrawLine(GILine* line, CGContextRef context, CGFloat offset, CGFloa
...
+      if (slashRange.location != NSNotFound) {
+        upstreamBranchNameRange.location += NSMaxRange(slashRange);
+        upstreamBranchNameRange.length -= NSMaxRange(slashRange);
+      }
+      _AppendAttributedString((CFMutableAttributedStringRef)multilineTitle, [NSString stringWithFormat:@"⬅ %@\n", upstreamName], multilineTitleAttributes);
+      [multilineTitle addAttribute:NSForegroundColorAttributeName value:darkColor range:upstreamBranchNameRange];
+
```
swisspol: Please remove blank line so that it matches better the coding style of GitUp.

GitUpKit/Interface/GIGraphView.m
```diff
@@ -860,112 +860,164 @@ static void _DrawLine(GILine* line, CGContextRef context, CGFloat offset, CGFloa
...
-    if (index) {
-      [label appendString:@", "];
-    }
-    NSString* upstream = localBranch.upstream.name;
-    if (upstream) {
-      [label appendFormat:@"%@ ◀ %@", localBranch.name, upstream];
+    
+    NSString* upstreamName = localBranch.upstream.name;
+    if (upstreamName) {
+
```
swisspol: Please remove blank line so that it matches better the coding style of GitUp.

shpakovski: 👌

GitUpKit/Interface/GIGraphView.m
shpakovski: 👻

GitUpKit/Interface/GIGraphView.m
shpakovski: I removed `> 0`, but `\n` is needed by design, this is what makes separator meaningful.

GitUpKit/Interface/GIGraphView.m
shpakovski: The final “separator” makes visual connection more evident.

GitUpKit/Interface/GIGraphView.m
shpakovski: So I need the branch name without `origin/` part from the `localBranch.upstream`. Which API should I use? Thanks in advance!

GitUpKit/Interface/GIGraphView.m
shpakovski: Done.

GitUpKit/Interface/GIGraphView.m
shpakovski: I to make it pixel-perfect, but depending on the font size looks better. Fixed.

shpakovski:
>When the PR complete, just flatten everything into a single commit and I'll merge it.

Alright, let me know when we are done.

GitUpKit/Interface/GIGraphView.m
shpakovski: Condition `stringRange.length == 1` is true only for `\n` which means separator line.

GitUpKit/Interface/GIGraphView.m
shpakovski: Correct, this should be less confusing. Thanks!

GitUpKit/Interface/GIGraphView.m
swisspol: Great, thanks much!

GitUpKit/Interface/GIGraphView.m
swisspol: Something like this:
```
GCBranch* upstream = localBranch.upstream;
if ([upstream isKindOfClass:[GCRemoteBranch class]]) {
  // upstream is a remote branch, use [(GCRemoteBranch*)upstream branchName] to get the branch name without the remote name part
} else {
  // upstream is a local branch
}
```
Note how such code is quite better as it clearly expresses its intent (i.e. distinguishing between local and remote upstreams), instead of doing it indirectly by manipulating branch names.

shpakovski:
Should be fixed now, please review 😉

swisspol:
I don't understand: my comments disappeared and I only see 1 commit again. Did you force push?

shpakovski:
Yes, sorry. The comments were so small, that I decided it is worth to unite the latest changes.

swisspol:
The other unfortunate side effect of a force push is that I don't think what you have changed since the last time I reviewed :disappointed:

shpakovski:
My bad, sorry… Let's agree, I don't force push until you say it is ready.

swisspol:
I reviewed the code again, I think it's good now! However after further testing in the app, I found 2 issues:

1) Tags are shown as `[tag]` but the brackets use the color of the line instead of the color of the text:

image

2) With the current implementation labels always occupy 2 lines and things are laid out so they can't possibly overlap:

image

This doesn't work anymore with the new design:

image

swisspol:
I also tweaked the rendering a bit to make it sharper and a bit darker (please apply to your branch with `patch -p1 < {PATCH_FILE}`):
```
diff --git a/GitUpKit/Interface/GIGraphView.m b/GitUpKit/Interface/GIGraphView.m
index b986eaa..30f7720 100644
--- a/GitUpKit/Interface/GIGraphView.m
+++ b/GitUpKit/Interface/GIGraphView.m
@@ -876,14 +876,14 @@ static void _DrawBranchTitle(CGContextRef context, CGFloat x, CGFloat y, NSColor
     multilineTitleAttributes[NSParagraphStyleAttributeName] = style;
     [style release];
   }
-  multilineTitleAttributes[NSForegroundColorAttributeName] = [color shadowWithLevel:0.25];
+  multilineTitleAttributes[NSForegroundColorAttributeName] = [color shadowWithLevel:0.2];
   
   // Cache bold font and calculate darker color for building multiline string
   static CTFontRef boldFont = NULL;
   if (boldFont == NULL) {
     boldFont = CTFontCreateUIFontForLanguage(kCTFontUIFontEmphasizedSystem, 13.0, CFSTR("en-US"));
   }
-  NSColor* darkColor = [color shadowWithLevel:0.7];
+  NSColor* darkColor = [NSColor colorWithDeviceWhite:0.2 alpha:1.0];
   
   // Start new attributed string for the branch title
   NSMutableAttributedString* multilineTitle = [[NSMutableAttributedString alloc] initWithString:@""];
@@ -953,9 +953,9 @@ static void _DrawBranchTitle(CGContextRef context, CGFloat x, CGFloat y, NSColor
   CGContextSaveGState(context);
   CGContextTranslateCTM(context, x, y);
   CGContextRotateCTM(context, 45.0 / 180.0 * M_PI);
-  CGContextSetLineWidth(context, 0.5);
-  
-  // Draw text
+
+  // Draw text and separators
+  NSColor* separatorColor = [color colorWithAlphaComponent:0.6];
   CGFloat lastLineWidth = 0.0;
   CFArrayRef lines = CTFrameGetLines(frame);
   for (CFIndex i = 0, count = CFArrayGetCount(lines); i < count; ++i) {
@@ -973,11 +973,11 @@ static void _DrawBranchTitle(CGContextRef context, CGFloat x, CGFloat y, NSColor
     // Draw separator in case of new line which is guaranteed by the building algorithm
     CFRange stringRange = CTLineGetStringRange(line);
     if (stringRange.length == 1) {
-      CGRect underlineRect = CGRectMake(origin.x - 1.0, origin.y - 1.0, lastLineWidth + ascent, ascent - descent);
-      CGContextMoveToPoint(context, underlineRect.origin.x, underlineRect.origin.y);
-      CGContextAddLineToPoint(context, underlineRect.origin.x + underlineRect.size.height, underlineRect.origin.y + underlineRect.size.height);
+      CGRect underlineRect = CGRectMake(floor(origin.x - 1.0), floor(origin.y - 1.0), ceil(lastLineWidth + ascent - descent), ceil(ascent - descent - 1.0));
+      CGContextMoveToPoint(context, underlineRect.origin.x + 0.5, underlineRect.origin.y);
+      CGContextAddLineToPoint(context, underlineRect.origin.x + underlineRect.size.height + 0.5, underlineRect.origin.y + underlineRect.size.height);
       CGContextAddLineToPoint(context, underlineRect.origin.x + underlineRect.size.width, underlineRect.origin.y + underlineRect.size.height);
-      CGContextSetStrokeColorWithColor(context, [color colorWithAlphaComponent:0.5].CGColor);
+      CGContextSetStrokeColorWithColor(context, separatorColor.CGColor);
       CGContextStrokePath(context);
       continue;
     }
```

Before:

image

After:

image

shpakovski:
> Please apply to your branch with patch…

Done. Thanks!

> …the brackets use the color of the line instead of the color of the text.

This is by design. Originally, these brackets were almost transparent…

> With the current implementation labels always occupy 2 lines… This doesn't work anymore…

Correct, it does not work in some cases, but there is a tradeoff. In the current one-line implementation I could not see some titles at all, this is how the PR was born. Also, original 12.0 font size worked better and reduced the risk of overlapping.

Sorry for the slow reply.

P. S. I did not rebase on `upstream` not `--force push` to avoid erasing any comments.

tomaz:
This change makes branches much more readable for me! Regarding labels overlapping; it can be more severe than than above screenshot. Branch labels with remotes need to use additional height:

gitup

shpakovski:
> Branch labels with remotes need to use additional height

Thank you for the feedback Tom! Could you please be more specific? What you mean by ‘additional height’?

tomaz:
Check dotted light green line: looks like overall height taken for that column is as if each of the texts would have single line. But top two texts occupy 2 lines each, so the text ovelaps "vadim/new-branch-titles" column. If text is composed from 2 lines (usually when there's remote), then it should use additional height to prevent it from overlapping. Hope it makes sense :)

Didn't check source code (yet), so not sure how it's handled behind the scenes. Looks like measuring thing...

shpakovski:
> If text is composed from 2 lines (usually when there's remote), then it should use additional height to prevent it from overlapping.

But in this case it may overlap with titles from the left branch. Sorry if I misunderstand your idea.

And again, I cannot get which part should get additional height… Graphical representation would help very much 😊 Thanks in advance!

tomaz:
See my screenshot above. Here it is again with an arrows pointing to problematic areas. And yes, that's exactly what happens: it overlaps the brach on the left side. What I propose is to take into account branch titles that spread multiple lines and have enough space to render the whole text without overlapping other branches. This might be solved by either take more space vertically or horizontally, whichever is more preferred.

gitup

shpakovski:
Thanks, now we have terms to talk in 🙃 So I'm curious about that vertical space. How exactly do you add it to avoid overlapping? There is no guarantee that left branch does not use different “combination” of single and double lines and vertical spaces which still lead to overlapping with the next branch. What do I miss, sorry?

Horizontal gap is a reliable solution which could work, but it would influence visual map appearance too much.

We could draw branches in reverse order and remember previous bounds to limit the drawing area, but this would require much more calculations. Is it worth it? 😦

tomaz:
I know, both have pros and cons... Personally I'd prefer variable horizontal gaps with predefined max width, then truncate text if it gets larger than that. It's probably simpler to implement too.

shpakovski:
> Personally I'd prefer variable horizontal gaps with predefined max width

Me too. BTW we may avoid additional gap when it is not needed e.g. the first branch on your snapshot fits nicely even though it is overlapped 😆

Pierre-Olivier, what do you think about this solution?

swisspol:
I like your UI change and I'd really like to get this in but right now it's not possible. This change introduces a UI regression and this is not acceptable :disappointed: . Properly truncating branch names and tags if too long, the current behavior, is one thing, but having rendering bugs is completely different:

image

You can't change the graph itself, only the labels. So yeah, it seems you'll have to draw labels on branches from right to left and remember the "bounding area" of the labels on the previous branch, then truncate text as needed (which CoreText can do for you).

Whatever you do, don't force push though :wink: . Since this PR is going back from review state to WIP state, it's OK to have many small commits in it.

shpakovski:
> It seems you'll have to draw labels on branches from right to left and remember the "bounding area" of the labels on the previous branch.

Done. Please verify.

swisspol:
Great work! One minor thing, then you can rebase and flatten everything (but don't close and re-open the PR :wink: ) and I'll merge this in:

Add 1 extra pixel between the lines: "new-branch-title" and "patch/new-branch-title" are a little close.

image

shpakovski:
Thanks a lot for your patience! Please review again. Is it correct? 😉

swisspol:
Yep, thanks again, this is great! It certainly took quite a bit of back and forth but it was definitely worth it IMO. The final implementation is much better :smile:

shpakovski:
Indeed, cheers! Thanks again for GitUp 👍

salutis:
Fantastic work, @shpakovski and @swisspol. Thanks!

I found two issues with this:
http://forums.gitup.co/t/two-bugs-regarding-new-branch-titles/403

shpakovski:
Thank you @salutis! I live on Retina, and find those fonts super-readable, even with System Preferences / Displays / Resolution / Scaled / More Space.

>Labels are not entirely visible.
That's true… @swisspol Calculating top space required for all titles is too hard, but maybe it is possible to allow drawing above the top edge, so that content is visible when you “ruber-band” (keep scrolling to show more contents)?

swisspol:
Guys, let's discuss this in the forum topic: this PR is closed and having things in 2 places is not great either :smile:

PR #25, opened 2015-12-07
Safer remote operations

tomaz:
I find it too easy to accidentally break remote (force push, delete branches etc.). This PR adds small tweak for such alerts so that confirm button is not marked as default. Instead user needs to either use mouse or tab+space to confirm, so gets more time to make sure that's what they want. This is off by default, so existing users are not affected, needs to be explicitly enabled.

Changes:

- Added `kGIAlertType_Danger` type which is used for potential data loss changes (mainly remote).
- Added `GIViewControllerUserDefaultKey_SaferRemoteConfirmations` user default which enables/disabled this behavior (default = false).
- Added preferences UI in advanced settings that changes above user default.

As far as UI change goes, it's convenient but can remain hidden preference with no UI as far as I'm concerned, set by handling `defaults` in Terminal - it's not something that will be tweaked daily. That's why I split PR into 2 commits; if UI change is not desired, I can remove last one, but having an option for preventing too easy data loss on remote is something I find crucial...

The reasoning behind this is: all GitUp alerts are confirmable simply by pressing enter. While all dangerous operations are guarded behind additional alert and use stop icon to differentiate, the behavior is still the same - just press enter to confirm. This behavior makes GitUp different from git command line - all "force" operations need to be explicitly invoked by user (by adding `--force` flag for example). Thus it's too easy to accidentally hit enter as a matter of habit when alert appears. This is additionally dangerous when user is expecting there will be 2 alerts, but there's only one. For example delete remote branch when there's no local counterpart - GitUp shows local and remote branches together at the top of the view, using the same font and color, the only difference is presence of slash symbol for remotes, and it's not as pronounced in angled as it would be in horizontal text (I find this aspect better with https://github.com/git-up/GitUp/pull/23).

### DIFF
--- a/GitUpKit/Utilities/GIViewController.m
+++ b/GitUpKit/Utilities/GIViewController.m
@@ -265,7 +265,10 @@
     alert.messageText = title;
     alert.informativeText = message;
     alert.showsSuppressionButton = key ? YES : NO;
-    [alert addButtonWithTitle:button];
+    NSButton* defaultButton = [alert addButtonWithTitle:button];
+    if (type == kGIAlertType_Danger) {
+      defaultButton.keyEquivalent = @"";
+    }
     [alert addButtonWithTitle:NSLocalizedString(@"Cancel", nil)];
     [self presentAlert:alert completionHandler:^(NSModalResponse returnCode) {
       
--- a/GitUpKit/Views/GIMapViewController+Operations.m
+++ b/GitUpKit/Views/GIMapViewController+Operations.m
@@ -31,6 +31,7 @@
 #define kUserDefaultsKey_SkipPushBranchWarning kUserDefaultsPrefix "SkipPushBranchWarning"
 #define kUserDefaultsKey_SkipPushLocalBranchToRemoteWarning kUserDefaultsPrefix "SkipPushLocalBranchToRemoteWarning"
 #define kUserDefaultsKey_SkipFetchRemoteBranchesWarning kUserDefaultsPrefix "SkipFetchRemoteBranchesWarning"
+#define kUserDefaultsKey_AllowReturnKeyForDangerousRemoteOperations kUserDefaultsPrefix "AllowReturnKeyForDangerousRemoteOperations"
 
 @interface GIMapViewController (Internal)
 - (void)_promptForCommitMessage:(NSString*)message withTitle:(NSString*)title button:(NSString*)button block:(void (^)(NSString* message))block;
@@ -38,6 +39,10 @@
 
 static inline NSString* _CleanedUpCommitMessage(NSString* message) {
   return [message stringByTrimmingCharactersInSet:[NSCharacterSet newlineCharacterSet]];
+}
+
+static inline GIAlertType _AlertTypeForDangerousRemoteOperations() {
+  return ([[NSUserDefaults standardUserDefaults] boolForKey:kUserDefaultsKey_AllowReturnKeyForDangerousRemoteOperations] ? kGIAlertType_Stop : kGIAlertType_Danger);
 }
 
 @implementation GIMapViewController (Operations)
@@ -456,7 +461,7 @@ static inline NSString* _CleanedUpCommitMessage(NSString* message) {
     
   }]) {
     if ([upstream isKindOfClass:[GCHistoryRemoteBranch class]]) {
-      [self confirmUserActionWithAlertType:kGIAlertType_Stop
+      [self confirmUserActionWithAlertType:_AlertTypeForDangerousRemoteOperations()
                                      title:[NSString stringWithFormat:NSLocalizedString(@"Do you also want to delete the upstream remote branch \"%@\" from its remote?", nil), upstream.name]
                                    message:NSLocalizedString(@"This action cannot be undone.", nil)
                                     button:NSLocalizedString(@"Delete Remote Branch", nil)
@@ -924,7 +929,7 @@ static inline NSString* _CleanedUpCommitMessage(NSString* message) {
       }
       
     } else if (!force && [error.domain isEqualToString:GCErrorDomain] && (error.code == kGCErrorCode_NonFastForward)) {
-      [self confirmUserActionWithAlertType:kGIAlertType_Stop
+      [self confirmUserActionWithAlertType:_AlertTypeForDangerousRemoteOperations()
                                      title:[NSString stringWithFormat:NSLocalizedString(@"The branch \"%@\" could not be fast-forwarded on the remote \"%@\". Do you want to attempt to force push?", nil), branch.name, remote ? remote.name : upstreamRemote.name]
                                    message:NSLocalizedString(@"This action cannot be undone.", nil)
                                     button:NSLocalizedString(@"Force Push", nil)
@@ -978,7 +983,7 @@ static inline NSString* _CleanedUpCommitMessage(NSString* message) {
     if (success) {
       [self.windowController showOverlayWithStyle:kGIOverlayStyle_Informational format:NSLocalizedString(@"All branches were pushed to the remote \"%@\" successfully!", nil), remote.name];
     } else if ([error.domain isEqualToString:GCErrorDomain] && (error.code == kGCErrorCode_NonFastForward)) {
-      [self confirmUserActionWithAlertType:kGIAlertType_Stop
+      [self confirmUserActionWithAlertType:_AlertTypeForDangerousRemoteOperations()
                                      title:[NSString stringWithFormat:NSLocalizedString(@"Some branches could not be fast-forwarded on the remote \"%@\". Do you want to attempt to force push?", nil), remote.name]
                                    message:NSLocalizedString(@"This action cannot be undone.", nil)
                                     button:NSLocalizedString(@"Force Push", nil)
@@ -1000,7 +1005,7 @@ static inline NSString* _CleanedUpCommitMessage(NSString* message) {
   NSArray* remotes = [self.repository listRemotes:&localError];
   if (remotes.count <= 1) {
     GCRemote* remote = remotes[0];
-    [self confirmUserActionWithAlertType:kGIAlertType_Stop
+    [self confirmUserActionWithAlertType:_AlertTypeForDangerousRemoteOperations()
                                    title:[NSString stringWithFormat:NSLocalizedString(@"Are you sure you want to push all branches to the remote \"%@\"?", nil), remote.name]
                                  message:NSLocalizedString(@"This action cannot be undone.", nil)
                                   button:NSLocalizedString(@"Push All Branches", nil)
@@ -1056,7 +1061,7 @@ static inline NSString* _CleanedUpCommitMessage(NSString* message) {
   NSArray* remotes = [self.repository listRemotes:&localError];
   if (remotes.count <= 1) {
     GCRemote* remote = remotes[0];
-    [self confirmUserActionWithAlertType:kGIAlertType_Stop
+    [self confirmUserActionWithAlertType:_AlertTypeForDangerousRemoteOperations()
                                    title:[NSString stringWithFormat:NSLocalizedString(@"Are you sure you want to push all tags to the remote \"%@\"?", nil), remote.name]
                                  message:NSLocalizedString(@"This action cannot be undone.", nil)
                                   button:NSLocalizedString(@"Push All Tags", nil)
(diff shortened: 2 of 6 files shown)


swisspol:
Thanks for sending this in. Please review first https://github.com/git-up/GitUp/blob/master/CONTRIBUTING.md.

This should be made of a couple commits:
- One to introduce a new `kGIAlertType_Danger` type and a different icon for it (not sure what)
- One to introduce the new user default in code and UI

What version of Xcode did you use? It seems to generate a huge diff for the XIB file while it's a simple change.

Finally please provide a screenshot of the updated preference panel.

tomaz:
I did read contributor guide.

Fixed indenting, sorry about those, I'm quite OCD about this, assumed Xcode did correct indenting as it all looked fine for me. Also fixed other code style issues.

Using Xcode 7.1.1. Checked the diff, it's mostly Xcode behind the scenes stuff throughout the file, that usually happens as soon as you open the XIB in newer Xcode. Perhaps splitting each view/window into its own XIB would help isolate changes for future. If you want me I can remove all unwanted changes manually and amend/rebase my PR.

There are 2 commits for this change - one for underlying behavior and one for UI change, mentioned this in my PR. Plus third one with code style fixes now.

gitup preferences

GitUpKit/Utilities/GIViewController.m
```diff
@@ -265,7 +267,12 @@ - (void)confirmUserActionWithAlertType:(GIAlertType)type
     alert.messageText = title;
     alert.informativeText = message;
     alert.showsSuppressionButton = key ? YES : NO;
-    [alert addButtonWithTitle:button];
+
```
swisspol: Remove blank line

GitUpKit/Utilities/GIViewController.m
```diff
@@ -265,7 +267,12 @@ - (void)confirmUserActionWithAlertType:(GIAlertType)type
     alert.messageText = title;
     alert.informativeText = message;
     alert.showsSuppressionButton = key ? YES : NO;
-    [alert addButtonWithTitle:button];
+		
+    NSButton* defaultButton = [alert addButtonWithTitle:button];
+    if ((type == kGIAlertType_Danger) && [[NSUserDefaults standardUserDefaults] boolForKey:GIViewControllerUserDefaultKey_SaferRemoteConfirmations]) {
+      defaultButton.keyEquivalent = @"";
+		}
+
```
swisspol: Remove blank line

swisspol:
Thinking more about the user default `GIViewControllerUserDefaultKey_AllowDangerous` is not a good fit.

It's on GIViewController, which has nothing to do with remote operations, so if the default is defined there, then it should be called `GIViewControllerUserDefaultKey_DisableDefaultButtonForDangerousAlerts` or something. But then this doesn't make much sense in the preferences.

Instead the default locality should match its intent: if it's about remote operations done in `GIMapViewController` then it must be defined there and used as such:

    [self confirmUserActionWithAlertType:([[NSUserDefaults ...] boolForKey...] ? kGIAlertType_Danger : kGIAlertType_Stop) ...

swisspol:
The user default behavior needs to be flipped too: default is OFF by default but that indicates the can't confirm with Return behavior.

The user default UI should go in the "Remotes:" section like:
```
Remotes: [ ] Allow confirming dangerous operations with Return key
         Automatically check...
```
However naming the default in code and figuring out a concise but still clear label is hard.

swisspol:
You can use this for the default:
```
#define kUserDefaultsKey_AllowReturnKeyForDangerousRemoteOperations kUserDefaultsPrefix "AllowReturnKeyForDangerousRemoteOperations"
```

swisspol:
That's the XIB problem BTW, you must have said yes:

image

Just say NO at the prompt and Xcode will barely modify the XIB.

tomaz:
There was no dialog, it just upgraded when I saved - it always works like this for me, perhaps there's a preference for this that I switched once and forgot :) Will update PR according to your comments.

tomaz:
About `#define kUserDefaultsKey_AllowReturnKeyForDangerousRemoteOperations kUserDefaultsPrefix "AllowReturnKeyForDangerousRemoteOperations"` - if original default behavior is to be preserved, then factory default needs to be set to YES. However if I want to have the #define alongside existing `GIMapViewController` keys, then it's private and can't be used in AppDelegate.

I'd rename the key to `PreventReturnKeyForDangerousRemoteOperations`, then it will be NO by default and no factory default setting will be needed to preserve previous behavior (that's how the other keys in GIMapViewController work and is cleanest solution IMO)

Other solutions:

- Have this key defined as `const NSString` with `extern` in header so it can be accessed in AppDelegate
- Have class method on `GIMapViewController` that will be called from `[AppDelegate initialize]` and will provide default values that can be injected to factory defaults together with the rest.

Please advise.

swisspol:
The default can switch to `NO` i.e. change the current behavior. That's OK: I'm not against the idea of making something safer, as long as there's an option to revert to previous behavior, which is the case here.

tomaz:
Ok, thanks.

One more thing: are you fine with adding convenience method to `GIMapViewController` that will return kGIAlertType and handle `([[NSUserDefaults ...] boolForKey:...] ? kGIAlertType_Danger : kGIAlertType_Stop)` internally? Will DRY the code and make it more readable. Alternatively C function can also be used.

swisspol:
That's reasonable, just make it a `static inline` and move to the top of the file - see `_CleanedUpCommitMessage()`.

tomaz:
Updated code & UI as per your comments (reverted previous UI commit so this one only includes actual changes, no Xcode 7 stuff).

Here's new screenshot:

gitup preferences

Let me know if there's anything else, otherwise I can flatten commits and rebase...

GitUpKit/Views/GIMapViewController+Operations.m
```diff
@@ -40,6 +41,10 @@ - (void)_promptForCommitMessage:(NSString*)message withTitle:(NSString*)title bu
   return [message stringByTrimmingCharactersInSet:[NSCharacterSet newlineCharacterSet]];
 }
 
+static inline GIAlertType _AlertTypeForDangerousRemoteOperations() {
+	return ([[NSUserDefaults standardUserDefaults] boolForKey:kUserDefaultsKey_AllowReturnKeyForDangerousRemoteOperations] ? kGIAlertType_Stop : kGIAlertType_Danger);
```
swisspol: Indenting looks wrong. Make sure to set Xcode indenting to 2 spaces in the project info.

GitUpKit/Utilities/GIViewController.m
```diff
@@ -265,7 +267,10 @@ - (void)confirmUserActionWithAlertType:(GIAlertType)type
     alert.messageText = title;
     alert.informativeText = message;
     alert.showsSuppressionButton = key ? YES : NO;
-    [alert addButtonWithTitle:button];
+    NSButton* defaultButton = [alert addButtonWithTitle:button];
+    if (type == kGIAlertType_Danger) {
+      defaultButton.keyEquivalent = @"";
+		}
```
swisspol: Indenting.

GitUpKit/Utilities/GIViewController.h
```diff
@@ -15,6 +15,8 @@
 
 #import "GIAppKit.h"
 
+extern NSString* const GIViewControllerUserDefaultKey_SaferRemoteConfirmations;
```
swisspol: Forgot to remove this.

swisspol:
Looks like you had to resize the preference window horizontally... The label is still not super clear. How about this (1 checkbox and then 1 text field underneath)?
```
[ ] Allow confirming dialogs for dangerous operations
    using Return key shortcut
```

tomaz:
Yes, had to resize the window, otherwise text wouldn't fit... Above suggestion is better and would fit 500pt, but still feels hard to grasp - had to read it couple times to "get it" and I'm the one working on the feature, so I can imagine it'll be an issue for others... (granted, I'm not native English speaker and it's been a long day here, so am not at the best :)

I try to keep checkboxes in preferences short, and instead add smaller explanation label below - similar to "Use simple mode" option on general pane. Unless you're sure about it, I'd sleep it over and see if I can find an alternative. Feel free to update here if you get other solutions.

Sorry for indentation, Xcode "fixed" couple lines on my behalf... Can't change preferences though, need to keep 4 char tabs for all of my client work. Feels too cumbersome to switch between settings as I switch between projects.

swisspol:
Sure, sleep over it :smile: 

You can set the indentation per project in Xcode:

image

tomaz:
Oh, right - completely forgot about this, thanks for tip. Strange it doesn't get committed...

shpakovski:
Sorry guys, my 2 cents. I truly believe that such preference should be hidden.

tomaz:
It is on the edge indeed... I was thinking in this direction as well, though I would implement some UI to access those, to make them more discoverable. Perhaps adding third preferences pane that would simply be a table view of grouped items (checkbox+description label). Adding new options there should be simple matter of adding the option in code - no XIB file changes should be required as those are more difficult to track...

I can implement such a system as part of this PR (or make this PR only include hidden preference and add UI changes in different one to keep them separated) - Pierre-Olivier how do you feel about this? What would the name and icon for this pane be then? I'm on the cheerful side personally, so "Hard Core" might fit (grinning face for icon??) :sweat_smile:

tomaz:
Otherwise here are couple ideas for the option in Advanced tab:

gitup preferences
gitup preferences 2
gitup preferences 3

shpakovski:
My version if you don't mind:
```
Remotes:  Automatically check  [ Every 15 minutes ]

          [  ]  Allow Return key for dangerous operations
		 
		  If checked, you can confirm remote operations
		  Force Push and Remove Branch using Return key.		 
```

tomaz:
It's short, like it. Not 100% about adding operations to description, it would require change if more operations are added in future.

shpakovski:
> Not 100% about adding operations to description

This is a hint which should prevent you from checking the box! 👻

tomaz:
:)

swisspol:
I think the copy should be:
```
[ ] Allow quick confirmation of dangerous operations
When enabled, even dangerous operations on remotes, like force push, can be confirmed by pressing the Return key in dialogs.
```

swisspol:
> It is on the edge indeed... I was thinking in this direction as well, though I would implement some UI to access those, to make them more discoverable. Perhaps adding third preferences pane that would simply be a table view of grouped items (checkbox+description label).

This would run contrary to the UX philosophy of GitUp: deceptively powerful while maintaining a simple UI. I don't want another a gazillion preferences like in Eclipse :smile: The other problem is that when you open that door of adding a custom option for a marginal number of users, it's a slippery slope.

This PR is slightly different because I understand the idea that dangerous remote operations should require extra confirmation. That said, for such operations there's an extra dialog you cannot skip, it's a big red icon and all, and in real life, GitUp has been out for 9+ months, is used by many developers, and this hasn't been a problem at all.

I'm borderline on this change, for the reasons above, but at the same time, GitUp goes great lengths to be very safe / undoable, so where it can't be, the UX must induce caution. Disabling the Return key would help in this case. Personally, I don't need this so I'd use the pref to restore the old behavior.

Anyway, as long as this PR is well contained and as small of a diff as possible, I'm OK.

tomaz:
Done, see screenshot.

Agree on keeping UI simple...

gitup preferences

swisspol:
That looks good, thanks. Please use this final copy which is a bit shorter, then rebase and flatten the PR, then I can merge it.

> This allows dangerous operations on remotes, like force push, to be confirmed by pressing the Return key in dialogs.

swisspol:
Any chance you could wrap this up soon, I'd to merge this in to release a new build.

tomaz:
I did - I changed the text + rebased & flattened PR, did I miss something?

swisspol:
Oh, I never received a notification from GitHub... Anyway, it's merged now, thanks!

tomaz:
np, thanks for bearing with me and merging - had busy end of week & weekend so didn't check much emails myself...

dmnd:
Thanks for your work @tomaz & @swisspol — just last Friday I accidentally deleted remote master :fearful:, so it's nice to see this update today!

:heart:

tomaz:
Thanks! Happy this effort is useful to community... :blush:

swisspol:

tomaz:
I agree

Agreed back then too, I think it was on gitup website, but may be wrong...

Cheers!

swisspol:
Ah right, I wasn't sure. Thanks much!

PR #26, opened 2015-12-09
Fix for popup menu crash when there is no upstream

shpakovski:


swisspol:
Thanks! What's the way to reproduce this?

swisspol:
Also if you browse the code, I'm not sure the APIs `-isEqualTo...` are supposed to accept `nil`. I'm not on my computer right now, but it's possible none do. If you think about it, it's not a valid GitUpKit API use to ask if an object is equal to nil. So that bug might have to be fixed at the call site. Browse the rest of the code to figure it out :smile:

douglashill:
Nice to see you’re on to this. Very easy to reproduce: just make a branch (it has to have no upstream) then select it and try to show the contextual menu.

shpakovski:
Thank you Douglas 👍

swisspol:
Thanks for the report. This has been fixed in da9ab5c7bb8566ed31d76a82c67c8f72adaf4d4a.

shpakovski:
Cool, thanks 👍

PR #27, opened 2015-12-09
Add 30% horizontal space between branch lines

shpakovski:
New format for branch titles is more compact, so adding some horizontal space does not affect layout too much, but makes it look cleaner.

### DIFF
--- a/GitUpKit/Interface/GIGraphView.m
+++ b/GitUpKit/Interface/GIGraphView.m
@@ -30,7 +30,7 @@
 #define __DEBUG_DESCENDANTS__ 0
 #define __DEBUG_ANCESTORS__ 0
 
-#define kSpacingX 30
+#define kSpacingX 40
 #define kSpacingY 30
 
 #define kMainLineWidth 8


swisspol:
Can you provide a screenshot of before and after?

I can't recall if there was a reason both X and Y spacing were the same value. I think it might be needed because of the way the branch paths are drawn with their lines at 0, 90 or 45 degrees. Changing this might actually break things.

shpakovski:
I played with different features, everything seems to be working!

Before:
before

After:
after

P. S. Sorry for the slow reply 😆

swisspol:
I'm not going to accept this PR, sorry. Even if the change is *trivial* in the code, I don't have the bandwidth to figure out what could break after the X spacing is different than the Y spacing.

Furthermore, I think this would reduce the information density in the map. The current one has plenty of white space already:

image

shpakovski:
Of course, no problem 👌 7-letter branch names are quite rare in my projects, so this must be a matter of style. Thank you for consideration anyway!

salutis:
+1

Looking at my app in GitUp, it definitely needs a bit more whitespace around:

screen shot 2016-01-12 at 11 25 10

shpakovski:
@salutis Exactly my case 😃 Sometimes every second label is cut with ellipsis.

swisspol:
The solution is not to add more space between branches, which negatively affects the information density of the entire map, but it's to improve how labels are laid out.

What should work is increasing the vertical space between labels so that they can never overlap. That's how it worked before. Play with both this spacing and the rotation angle (which was 40 and not 45 before IIRC).

Think about drawing diagonal lines on a grid: they can never intersect by definition.

swisspol:
Actually this should be a new PR anyway, this change is not the right one,

shpakovski:
I see what you mean, will think about this…

PR #29, opened 2016-02-26
Added support for Vim keystrokes while navigating commits.

akosma:
Usual keystrokes, nothing fancy: "h" to go left, "j" to go up, "k" to go down and "l" to go right.

### DIFF
--- a/GitUpKit/Interface/GIGraphView.m
+++ b/GitUpKit/Interface/GIGraphView.m
@@ -529,6 +529,42 @@ static const void* _associatedObjectDataKey = &_associatedObjectDataKey;
       return;
     
   }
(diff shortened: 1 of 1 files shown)


swisspol:
Only arrow keys and Esc can be used for navigation as all the other keys are reserved for shortcuts in the [Map view contextual menu](https://github.com/git-up/GitUp/wiki/Using-GitUp-Map-View).

akosma:
Precisely, you're not using neither H, J, K nor L, and this are quite common (vim, man pages, more, etc)

Anyway, thanks.

swisspol:
Not used _for now_.

PR #73, opened 2016-03-10
Always show welcome when clicking on dock icon

deyton:
Before this change, if the (x) button was clicked to close the welcome
window, then clicking on the app icon in the dock did nothing and there
was no way to show the welcome window again. Now, clicking on the dock
icon will show the welcome window if there are no other visible windows.

This is how Xcode behaves, which seems like a good standard to emulate
and something many GitUp users would expect.

### DIFF
--- a/GitUp/Application/AppDelegate.m
+++ b/GitUp/Application/AppDelegate.m
@@ -391,6 +391,14 @@
   return NO;
 }
 
+- (BOOL)applicationShouldHandleReopen:(NSApplication*)theApplication hasVisibleWindows:(BOOL)hasVisibleWindows {
+  if (!hasVisibleWindows) {
+    _allowWelcome = 1; // Always show welcome when clicking on dock icon
+    [self handleDocumentCountChanged];
+  }
+  return YES;
+}
+
 - (void)applicationDidBecomeActive:(NSNotification*)notification {
   if (_allowWelcome < 0) {
     _allowWelcome = 1;


swisspol:
IIRC the app use `_allowWelcome` to prevent this very behavior in `-applicationDidBecomeActive:`, so removing this state should achieve this, no?

deyton:
That was my first idea, but I found two issues.

1. It didn't work if GitUp didn't become inactive between closing the welcome window and clicking the dock icon. (That is, press (x) button then immediately click on the dock icon again).
2. After fixing #1, I noticed that the welcome window would open if GitUp became active by some other means (for instance, via ⌘-tab). Xcode doesn't do this so I thought it wasn't a good thing.

I've always found the welcome window behavior odd, and I think it was because it didn't behave the same as Xcode so that was my goal here.

deyton:
So guess it's not just Xcode. I just tried every app in my dock, and they all load a window when clicking on the dock icon, and don't load a window when switching to them via ⌘-tab, so I think the behavior change here is in probably in line with Mac apps in general.

An alternative that would result in similar behavior would be to terminate/quit GitUp when there are no open windows.

swisspol:
>  I just tried every app in my dock, and they all load a window when clicking on the dock icon

That's standard NSDocumentController behavior, but it's disabled in GitUp since it makes no sense to create an "Untitled" empty repo as a new empty document.

Xcode is in the same case as GitUp though so copying its behavior would make sense.

GitUp/Application/AppDelegate.m
```diff
@@ -391,6 +391,14 @@ - (BOOL)applicationShouldOpenUntitledFile:(NSApplication*)sender {
   return NO;
 }
 
+- (BOOL)applicationShouldHandleReopen:(NSApplication *)theApplication hasVisibleWindows:(BOOL)hasVisibleWindows {
```
swisspol: No space before `*`

swisspol:
Fix and flatten, then I'll merge. Thx.

GitUp/Application/AppDelegate.m
deyton: Oops! I didn't notice that bit of the style. Fixed.

swisspol:

deyton:

swisspol:
Thanks much!

PR #77, opened 2016-03-12
Allow ⌘-W to close welcome window via NSWindow subclass

deyton:
This enables both the File -> Close menu item and the ⌘-W keyboard
shortcut to close the welcome window. Windows without a titlebar close
button don't get this behavior for free, but since we have a button that
closes the window, it makes sense to add it.

Inspired by #73 to make the welcome window behave more like Xcode's


swisspol:
Accepting the GitUp CLA is required before this can be merged.

deyton:
I just added "I AGREE TO THE GITUP CONTRIBUTOR LICENSE AGREEMENT" to the description - is that all that's needed?

swisspol:
Yep, thanks!

PR #121, opened 2016-04-19
Make the view options for the map view persistent by storing them in the repository

JamesBucanek:
These commits store the various map view options in the local repository's `userInfo` property value collection.

The map options are initialized from the repository's property values during initialization, and updated whenever the user changes them.

The window restoration code in `GIMapViewController` was removed because it is now superfluous. This also eliminates the "dual reloading" issue mentioned in `-restoreStateWithCoder:`.

### DIFF
--- a/GitUpKit/Views/GIMapViewController.m
+++ b/GitUpKit/Views/GIMapViewController.m
@@ -25,13 +25,15 @@
 #import "GCHistory+Rewrite.h"
 #import "XLFacilityMacros.h"
 
-#define kRestorableStateKey_ShowVirtualTips @"showVirtualTips"
-#define kRestorableStateKey_HideTagTips @"hideTagTips"
-#define kRestorableStateKey_HideRemoteBranchTips @"hideRemoteBranchTips"
-#define kRestorableStateKey_HideStaleBranchTips @"hideStaleBranchTips"
+#define kPersistentViewStateKeyNamespace @"GIMapViewController_"
 
-#define kRestorableStateKey_ShowTagLabels @"showTagLabels"
-#define kRestorableStateKey_ShowBranchLabels @"showBranchLabels"
+#define kPersistentViewStateKey_HideVirtualTips kPersistentViewStateKeyNamespace @"HideVirtualTips"
+#define kPersistentViewStateKey_ShowTagTips kPersistentViewStateKeyNamespace @"ShowTagTips"
+#define kPersistentViewStateKey_ShowRemoteBranchTips kPersistentViewStateKeyNamespace @"ShowRemoteBranchTips"
+#define kPersistentViewStateKey_ShowStaleBranchTips kPersistentViewStateKeyNamespace @"ShowStaleBranchTips"
+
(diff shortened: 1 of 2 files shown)


swisspol:
No need for `kRepositoryUserInfoKey_ViewState`, nor BOOL convenience methods: just put all the keys directly in the repo user info.

Also please follow style guide - review existing code.

swisspol:
The first commit need to remove all the code related to the window restoration, while the 2nd one implements window settings persistency through the repo info.

JamesBucanek:
> The first commit need to remove all the code related to the window restoration, while the 2nd one implements window settings persistency through the repo info.

I tried to do that, but I couldn't cherrypick a set of lines from the diff (which mixed up the code getting added with the code getting deleted) that would result in something that would compile. l guess I could try throwing everything away and reapplying the changes by hand...

> No need for `kRepositoryUserInfoKey_ViewState` ... just put all the keys directly in the repo user info.

I was just trying to be neat, but they could easily go into `userInfo` as individual values.

> nor BOOL convenience methods:

I have to disagree. A lot of the values have non-zero defaults and you need to test to see when the value changes; without these methods you'll end up writing the same 6 lines of code over and over again, which is poor form.

> Also please follow style guide - review existing code.

I was trying. You'll have to tell me what I missed.

swisspol:
Just use one-liners:
```
_hidesTagTips = [[self.repository userInfoForKey:@"some_key"] boolValue];

- (void)setHidesTagTips:(BOOL)flag {
  if (flag != _hidesTagTips) {
    _hidesTagTips = flag;
    [self.repository setUserInfo:@(_hidesTagTips) forKey:@"some_key"];
    [self _reloadMap:YES];
  }
}
```

And for these settings that have a `YES` default instead of a `NO`, just negate what is written to / read from the user info.

JamesBucanek:
OK, the second version is ready.

GitUpKit/Views/GIMapViewController.m
```diff
@@ -25,13 +25,13 @@
 #import "GCHistory+Rewrite.h"
 #import "XLFacilityMacros.h"
 
-#define kRestorableStateKey_ShowVirtualTips @"showVirtualTips"
```
swisspol: It's best to have something like:
```
#define kPersistentViewStateKeyNamespace "GIMapViewController_"
#define kPersistentViewStateKey_ShowBranchLabels kPersistentViewStateKeyNamespace "ShowBranchLabels"
#define kPersistentViewStateKey_ShowTagTips kPersistentViewStateKeyNamespace "ShowTagTips"
```
And of course:
```
_hideTagTips = ![... kPersistentViewStateKey_ShowTagTips]
```

JamesBucanek:
Did you want me to try an undo this last commit, or just submit a third one?

swisspol:
You can just make a 3rd commit, merge with 2nd one and force-push.

JamesBucanek:
Third time's the charm...

swisspol:
Good work! Notice how the final diff is much simpler and elegant than the original one too 😄

JamesBucanek:
It's been fun. Thanks for letting me contribute.

Issue #74, opened 2016-03-11
GitHub app URL scheme


jankeesvw:
On GitHub.com you find link with "Open this in GitHub Desktop" they use an URL-Scheme like: `github-mac://openRepo/https://github.com/project/reponame?branch=branch-name`

Would be great if GitUp would also support these.

swisspol:
How would this work if both GitHub and GitUp are installed?

JanX2:
SourceTree has a checkbox that sets it to be the designated app for that URL scheme. At least until the GitHub app claims it for itself once again. ;)

swisspol:
Follow up in PR.

PR #138, opened 2016-05-05
Add GitHub Desktop button support

lucasderraugh:
Addresses #74

### DIFF
--- a/GitUp/Application/AppDelegate.m
+++ b/GitUp/Application/AppDelegate.m
@@ -346,6 +346,8 @@
   // Initialize Google Analytics
   [[GARawTracker sharedTracker] startWithTrackingID:@"UA-83409580-1"];
 #endif
+  
+  [[NSAppleEventManager sharedAppleEventManager] setEventHandler:self andSelector:@selector(_getUrl:withReplyEvent:) forEventClass:kInternetEventClass andEventID:kAEGetURL];
 }
 
 - (void)applicationDidFinishLaunching:(NSNotification*)notification {
@@ -420,6 +422,16 @@
   // Enable sudden termination
   [[NSProcessInfo processInfo] enableSuddenTermination];
 #endif
+}
+
+- (void)_getUrl:(NSAppleEventDescriptor*)event withReplyEvent:(NSAppleEventDescriptor*)replyEvent {
+  NSURL* url = [NSURL URLWithString:[event paramDescriptorForKeyword:keyDirectObject].stringValue];
+  BOOL isGitHubMacScheme = [url.scheme rangeOfString:@"github-mac" options:NSCaseInsensitiveSearch].location != NSNotFound;
+  BOOL isOpenRepoHost = [url.host rangeOfString:@"openRepo" options:NSCaseInsensitiveSearch].location != NSNotFound;
+  NSString* path = url.path.length ? [url.path substringFromIndex:1] : nil;
+  if (isGitHubMacScheme && isOpenRepoHost && path) {
+    [self _cloneRepositoryFromURLString:path];
+  }
 }
 
 - (BOOL)applicationShouldOpenUntitledFile:(NSApplication*)sender {
@@ -592,8 +604,8 @@ static CFDataRef _MessagePortCallBack(CFMessagePortRef local, SInt32 msgid, CFDa
   }
 }
 
-- (IBAction)cloneRepository:(id)sender {
-  _cloneURLTextField.stringValue = @"";
+- (void)_cloneRepositoryFromURLString:(NSString*)urlString {
+  _cloneURLTextField.stringValue = urlString;
   _cloneRecursiveButton.state = NSOnState;
   if ([NSApp runModalForWindow:_cloneWindow] && _cloneURLTextField.stringValue.length) {
     NSURL* url = GCURLFromGitURL(_cloneURLTextField.stringValue);
@@ -630,6 +642,10 @@ static CFDataRef _MessagePortCallBack(CFMessagePortRef local, SInt32 msgid, CFDa
       [NSApp presentError:MAKE_ERROR(@"Invalid Git repository URL")];
     }
   }
+}
+
+- (IBAction)cloneRepository:(id)sender {
+  [self _cloneRepositoryFromURLString:@""];
 }
 
 - (IBAction)dimissModal:(id)sender {
--- a/GitUp/Application/Info.plist
+++ b/GitUp/Application/Info.plist
@@ -33,6 +33,15 @@
 	<string>${BUNDLE_VERSION}</string>
 	<key>CFBundleShortVersionString</key>
 	<string>${BUNDLE_VERSION_STRING}</string>
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>github-mac</string>
+			</array>
+		</dict>
+	</array>
 	<key>LSMinimumSystemVersion</key>
 	<string>$(MACOSX_DEPLOYMENT_TARGET)</string>
 	<key>NSHumanReadableCopyright</key>


GitUp/Application/AppDelegate.m
```diff
@@ -411,6 +413,11 @@ - (void)applicationDidFinishLaunching:(NSNotification*)notification {
 #endif
 }
 
+- (void)getUrl:(NSAppleEventDescriptor*)event withReplyEvent:(NSAppleEventDescriptor*)replyEvent {
```
swisspol: Private methods must start with underscore

GitUp/Application/AppDelegate.m
```diff
@@ -573,8 +580,8 @@ - (IBAction)newRepository:(id)sender {
   }
 }
 
-- (IBAction)cloneRepository:(id)sender {
-  _cloneURLTextField.stringValue = @"";
+- (void)cloneRepositoryFromURLString:(NSString*)urlString {
```
swisspol: Same

swisspol:
Before merging this in, we would need a description of all the cases where both gitup and github mac are installed e.g. what happens if one is installef and then the other? What happens when this rolls out in a GitUp update? Does it break things for people who want GitHub Mac to open these links? How do you pick one app vs the other? Etc...

lucasderraugh:
From what I can tell, GitHub Desktop always seems to take precedence as long as it is installed. Unclear how that works, but it alters the default application for that URL scheme to be GitHub Desktop.

JanX2:
@lucasderraugh Yes. And it does this repeatedly. One of the reasons I almost never use it. I have to use RCDefaultApp time and again to fix this. Haven’t started GitHub Desktop in month. SourceTree is what I use and have assigned to the GitHub URL scheme.

lucasderraugh:
So is there something I should change about this implementation to get it approved?

GitUp/Application/Info.plist
```diff
@@ -33,6 +33,19 @@
 	<string>${BUNDLE_VERSION}</string>
 	<key>CFBundleShortVersionString</key>
 	<string>${BUNDLE_VERSION_STRING}</string>
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>CFBundleURLName</key>
```
swisspol: The `CFBundleURLName` key is not present in the GitHub Desktop Mac app, they only have this:
```
	<key>CFBundleURLTypes</key>
	<array>
		<dict>
			<key>CFBundleURLName</key>
			<string>com.github.GitHub</string>
			<key>CFBundleURLSchemes</key>
			<array>
				<string>github-mac</string>
			</array>
		</dict>
	</array>
```

According to https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/TP40009249-SW1 this key is not required, so let's remove it to match GitHub.

GitUp/Application/AppDelegate.m
```diff
@@ -411,6 +413,11 @@ - (void)applicationDidFinishLaunching:(NSNotification*)notification {
 #endif
 }
 
+- (void)_getUrl:(NSAppleEventDescriptor*)event withReplyEvent:(NSAppleEventDescriptor*)replyEvent {
+  NSString* url = [[event paramDescriptorForKeyword:keyDirectObject].stringValue stringByReplacingOccurrencesOfString:@"github-mac://openRepo/" withString:@""];
```
swisspol: This doesn't seem very clean: what's the exact string passed retrieved from the event?

GitUp/Application/AppDelegate.m
```diff
@@ -411,6 +413,11 @@ - (void)applicationDidFinishLaunching:(NSNotification*)notification {
...
+  [self _cloneRepositoryFromURLString:url];
```
swisspol: The URL must be verified to be valid before calling the function.

GitUp/Application/AppDelegate.m
```diff
@@ -613,6 +620,10 @@ - (IBAction)cloneRepository:(id)sender {
   }
 }
 
+- (IBAction)cloneRepository:(id)sender {
+  [self _cloneRepositoryFromURLString:@""];
```
swisspol: That's not going to work...

lucasderraugh: Why would that not work? It's the equivalent of what was there before, substituting in @"" for the urlString param passing in @"".

swisspol: Oh, I thought your intent was to immediately prompt where to clone and clone, not show the repo URL dialog. I guess it makes sense like this.

lucasderraugh: My thought was there was still the option of whether the user wants to pull in submodules. I could change it to just assume yes if you'd prefer.

swisspol: That's fine. I haven't worked on this UI in forever, so I don't remember all the details 😉

GitUp/Application/AppDelegate.m
lucasderraugh: I mostly just took the info from here: https://gist.github.com/mralexgray/3915592
I'll see if I can simplify the call as I do agree it's sort of convoluted.

GitUp/Application/Info.plist
lucasderraugh: Did you mean the `CFBundleTypeRole`?

GitUp/Application/AppDelegate.m
lucasderraugh: From the header file:
```
// Set or remove a handler for a specific kind of Apple Event.  The handler method should have the same signature as:
// - (void)handleAppleEvent:(NSAppleEventDescriptor *)event withReplyEvent:(NSAppleEventDescriptor *)replyEvent;
```
I should probably keep it as it is.

GitUp/Application/AppDelegate.m
lucasderraugh: Is a check for the correct prefix good enough?

swisspol: I don't know. I have no idea what the format of these URLs is 😄

lucasderraugh: The url for this repo is github-mac://openRepo/https://github.com/git-up/GitUp. Essentially I can just check to see that the prefix is 'github-mac://openRepo/' if that's good enough?

swisspol: Yeah, but in a case insensitive way.

lucasderraugh: Should I also replace the prefix string case insensitively?

GitUp/Application/AppDelegate.m
```diff
@@ -335,6 +335,8 @@ - (void)applicationWillFinishLaunching:(NSNotification*)notification {
   [Crashlytics startWithAPIKey:@"946dcb56f0db1be8f0c52df6e7d392202a2cc9b2"];
   [[Crashlytics sharedInstance] setDelegate:self];
 #endif
+  
+  [[NSAppleEventManager sharedAppleEventManager] setEventHandler:self andSelector:@selector(_getUrl:withReplyEvent:) forEventClass:kInternetEventClass andEventID:kAEGetURL];
```
swisspol: Any reason why this is not in `-applicationDidFinishLaunching:`?

GitUp/Application/Info.plist
swisspol: I mean `CFBundleURLName`: just remove it since it's not required and GitHub Mac doesn't have it either.

GitUp/Application/AppDelegate.m
lucasderraugh: Ya, it wasn't registering early enough if I placed the code there in cases where the application wasn't already open.

GitUp/Application/AppDelegate.m
```diff
@@ -411,6 +413,14 @@ - (void)applicationDidFinishLaunching:(NSNotification*)notification {
 #endif
 }
 
+- (void)_getUrl:(NSAppleEventDescriptor*)event withReplyEvent:(NSAppleEventDescriptor*)replyEvent {
+  NSString* urlString = [event paramDescriptorForKeyword:keyDirectObject].stringValue;
+  NSString* githubPrefix = @"github-mac://openRepo/";
+  if ([urlString hasPrefix:githubPrefix]) {
+    [self _cloneRepositoryFromURLString:[urlString stringByReplacingOccurrencesOfString:githubPrefix withString:@"" options:NSCaseInsensitiveSearch range:NSMakeRange(0, urlString.length)]];
```
swisspol: This doesn't make sense: `hasPrefix:` is case sensitive, so doing a case insensitive search and replace afterwards doesn't add anything.

lucasderraugh: My bad, will change.

GitUp/Application/Info.plist
lucasderraugh: Confused, from your example above, they clearly have:
```
<key>CFBundleURLName</key>
<string>com.github.GitHub</string>
```

GitUp/Application/AppDelegate.m
swisspol: What does that mean in practice?

lucasderraugh: It means if I put it in `-applicationDidFinishLaunching:` it wouldn't trigger any event if I GitUp wasn't open yet.

GitUp/Application/AppDelegate.m
swisspol: If this thing is supposed to be a URL, use an actual `NSURL` for it, then case-insensitive compare the scheme and host and make sure the path is not empty. Finally pass the path to `_cloneRepositoryFromURLString:`

jankeesvw:
I would love this feature! ❤️

jankeesvw:
@swisspol can this be merged? Would save me a ton of time! ❤️

lucasderraugh:

swisspol:
Thanks

PR #147, opened 2016-05-19
Keep table position when dropping a stash

lucasderraugh:
For example, when you select the last row and then delete it, your selection goes all the way to the top of the table. This change means that if you select the last row and delete it, it will go to the row next up from the bottom.

### DIFF
--- a/GitUpKit/Views/GIStashListViewController.m
+++ b/GitUpKit/Views/GIStashListViewController.m
@@ -307,6 +307,9 @@
       [self.undoManager setActionName:NSLocalizedString(@"Drop Stash", nil)];
       [[self.undoManager prepareWithInvocationTarget:self] _undoDropStashWithPreviousState:currentState ignore:NO];  // TODO: We should really use the built-in undo mechanism from GCLiveRepository
       [self.repository notifyRepositoryChanged];
+      if (row > 0) {
+        [_tableView selectRowIndexes:[NSIndexSet indexSetWithIndex:(row - 1)] byExtendingSelection:NO];
+      }
     } else {
       [self presentError:error];
     }


swisspol:
What if there's only one row?

lucasderraugh:
It doesn't do anything. As long as the range of the NSIndexSet is within NSUInteger bounds it doesn't do anything. It just won't select anything in the table. The alternative is to just check if row == 0 then don't do anything but I figured this was slightly more compact. Let me know if it should be changed to check on that instead.

swisspol:
It's not about `if row == 0` but `if number of rows == 1` pre-deletion.

lucasderraugh:
I think we're talking about the same thing. If you have only 1 item in the table which is selected, then you delete that you are on `row == 0`. Same thing applies.

swisspol:
I was only pointing out the check `row == 0` is not valid since it doesn't imply `number of rows == 1`

lucasderraugh:
True. I just considered the edge condition to be when you went out of bounds :)

GitUpKit/Views/GIStashListViewController.m
```diff
@@ -307,6 +307,7 @@ - (IBAction)dropStash:(id)sender {
       [self.undoManager setActionName:NSLocalizedString(@"Drop Stash", nil)];
       [[self.undoManager prepareWithInvocationTarget:self] _undoDropStashWithPreviousState:currentState ignore:NO];  // TODO: We should really use the built-in undo mechanism from GCLiveRepository
       [self.repository notifyRepositoryChanged];
+      [_tableView selectRowIndexes:[NSIndexSet indexSetWithIndex:MAX(0, row - 1)] byExtendingSelection:NO];
```
swisspol: I would prefer this which explicitly handles the edge cases vs relying on "patching" them afterwards with `MAX()` or relaying on `-selectRowIndexes: byExtendingSelection:` being a no-op if the table is empty (I'm assuming here that `NSTableView` keeps the 0th row selected if deleting row 0):
```
if (row > 0) {
  [_tableView selectRowIndexes:[NSIndexSet indexSetWithIndex:(row - 1) byExtendingSelection:NO];
}
```

lucasderraugh:
You're right, relying on that behavior is a little risky. Updated.

swisspol:

lucasderraugh:
I agree to the contributor's license agreement

PR #172, opened 2016-07-18
Resolve name conflicts caused by symlinks while moving items to trash

nikolaykasyanov:
Without the fix "Discard All" action could fail in the middle if there's a symlink in Trash with the same name as one of the items being discarded.

This scenario is pretty common when working with CocoaPods (iOS/Mac dependency management tool).

I'm not sure if this is a good solution performance-wise. Another option is to use`stat()` instead of `-fileExistsAtPath:`.

`-fileExistsAtPath:` checklist:

- [x] https://github.com/git-up/GitUp/blob/59ab927d0a3f86671ce5045ab5c47131bba4a199/GitUp/Application/AppDelegate.m#L563 — switched to `-fileOrSymlinkExistsAtPath:`
- [x] https://github.com/git-up/GitUp/blob/59ab927d0a3f86671ce5045ab5c47131bba4a199/GitUp/Application/AppDelegate.m#L594 — switched to `-fileOrSymlinkExistsAtPath:`
- [x] https://github.com/git-up/GitUp/blob/59ab927d0a3f86671ce5045ab5c47131bba4a199/GitUpKit/Core/GCLiveRepository.m#L524 — snapshot loading, files aren't supposed to be symlink
- [ ] https://github.com/git-up/GitUp/blob/59ab927d0a3f86671ce5045ab5c47131bba4a199/GitUpKit/Extensions/GCRepository+Index.m#L119 — I don't know how does conflict resolving work for symlinks.
- [ ] https://github.com/git-up/GitUp/blob/59ab927d0a3f86671ce5045ab5c47131bba4a199/GitUpKit/Extensions/GCRepository+Utilities.m#L144 — I'm not sure how to handle moving of symlinks.
- [x] https://github.com/git-up/GitUp/blob/59ab927d0a3f86671ce5045ab5c47131bba4a199/GitUpKit/Extensions/GCRepository+Utilities.m#L515
- [x] https://github.com/git-up/GitUp/blob/59ab927d0a3f86671ce5045ab5c47131bba4a199/GitUpKit/Core/GCRepository.m#L208 — private app data path getter, symlinks aren't expected
- [ ] https://github.com/git-up/GitUp/blob/59ab927d0a3f86671ce5045ab5c47131bba4a199/GitUpKit/Core/GCSubmodule.m#L158 — the check is faulty if there's a symlink at the path of submodule, but I've found that `-[GCSubmodule checkSubmoduleInitialized:error:]` also doesn't like symlinks and decided not to touch it this check at all.
- [x] https://github.com/git-up/GitUp/blob/59ab927d0a3f86671ce5045ab5c47131bba4a199/GitUpKit/Utilities/GIViewController+Utilities.m#L142 — broken symlinks are staged & committed.

### DIFF
--- a/GitUp/Application/AppDelegate.m
+++ b/GitUp/Application/AppDelegate.m
@@ -562,7 +562,7 @@ static CFDataRef _MessagePortCallBack(CFMessagePortRef local, SInt32 msgid, CFDa
   if ([savePanel runModal] == NSFileHandlingPanelOKButton) {
     NSString* path = savePanel.URL.path;
     NSError* error;
-    if (![[NSFileManager defaultManager] fileExistsAtPath:path] || [[NSFileManager defaultManager] moveItemAtPathToTrash:path error:&error]) {
+    if (![[NSFileManager defaultManager] fileExistsAtPath:path followLastSymlink:NO] || [[NSFileManager defaultManager] moveItemAtPathToTrash:path error:&error]) {
       GCRepository* repository = [[GCRepository alloc] initWithNewLocalRepository:path bare:NO error:&error];
       if (repository) {
         [self _openRepositoryWithURL:[NSURL fileURLWithPath:repository.workingDirectoryPath] withCloneMode:kCloneMode_None windowModeID:NSNotFound];
@@ -593,7 +593,7 @@ static CFDataRef _MessagePortCallBack(CFMessagePortRef local, SInt32 msgid, CFDa
       if ([savePanel runModal] == NSFileHandlingPanelOKButton) {
         NSString* path = savePanel.URL.path;
         NSError* error;
-        if (![[NSFileManager defaultManager] fileExistsAtPath:path] || [[NSFileManager defaultManager] moveItemAtPathToTrash:path error:&error]) {
+        if (![[NSFileManager defaultManager] fileExistsAtPath:path followLastSymlink:NO] || [[NSFileManager defaultManager] moveItemAtPathToTrash:path error:&error]) {
           GCRepository* repository = [[GCRepository alloc] initWithNewLocalRepository:path bare:NO error:&error];
           if (repository) {
             if ([repository addRemoteWithName:@"origin" url:url error:&error]) {
--- a/GitUpKit/Core/GCFoundation.h
+++ b/GitUpKit/Core/GCFoundation.h
@@ -15,10 +15,12 @@
 
 #import <Foundation/Foundation.h>
 
+@interface NSFileManager (GCFoundation)
+- (BOOL)fileExistsAtPath:(NSString*)path followLastSymlink:(BOOL)followLastSymlink;
+
 #if !TARGET_OS_IPHONE
 
-@interface NSFileManager (GCFoundation)
 - (BOOL)moveItemAtPathToTrash:(NSString*)path error:(NSError**)error;
-@end
 
 #endif
+@end
--- a/GitUpKit/Core/GCFoundation.m
+++ b/GitUpKit/Core/GCFoundation.m
@@ -19,9 +19,24 @@
 
 #import "GCPrivate.h"
 
-#if !TARGET_OS_IPHONE
-
 @implementation NSFileManager (GCFoundation)
+
+- (BOOL)fileExistsAtPath:(NSString*)path followLastSymlink:(BOOL)followLastSymlink {
+  NSDictionary* pathAttributes = [self attributesOfItemAtPath:path error:NULL];
+
+  if (followLastSymlink && [[pathAttributes fileType] isEqualToString:NSFileTypeSymbolicLink]) {
+    path = [self destinationOfSymbolicLinkAtPath:path error:NULL];
+    if (!path) {
+      return NO;
+    }
+
+    return [self fileExistsAtPath:path followLastSymlink:YES];
+  }
+
+  return pathAttributes != nil;
+}
+
+#if !TARGET_OS_IPHONE
 
 - (BOOL)moveItemAtPathToTrash:(NSString*)path error:(NSError**)error {
   NSString* trashPath = [NSSearchPathForDirectoriesInDomains(NSTrashDirectory, NSUserDomainMask, YES) firstObject];
@@ -33,12 +48,12 @@
   NSString* name = [path.lastPathComponent stringByDeletingPathExtension];
   NSString* destinationPath = [trashPath stringByAppendingPathComponent:[name stringByAppendingPathExtension:extension]];
   NSUInteger counter = 0;
-  while ([self fileExistsAtPath:destinationPath]) {
+  while ([self fileExistsAtPath:destinationPath followLastSymlink:NO]) {
     destinationPath = [trashPath stringByAppendingPathComponent:[[NSString stringWithFormat:@"%@ (%lu)", name, ++counter] stringByAppendingPathExtension:extension]];
   }
   return [self moveItemAtPath:path toPath:destinationPath error:error];
 }
 
-@end
-
 #endif
+
+@end
--- a/GitUpKit/Core/GCFoundation-Tests.m
+++ b/GitUpKit/Core/GCFoundation-Tests.m
@@ -1,0 +1,168 @@
+//  Copyright (C) 2015 Pierre-Olivier Latour <info@pol-online.net>
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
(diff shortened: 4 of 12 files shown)


swisspol:
So it seems the problem is that `NSFileManager fileExistsAtPath` follows symlinks. Just replace it with `attributesOfItemAtPath` which doesn't. We just care if anything exists at that path, it does not matter what type it is.

It's also worth checking if `fileExistsAtPath` is used anywhere else in the code.

swisspol:
Also please review https://github.com/git-up/GitUp/blob/master/CONTRIBUTING.md and indicate that you agree to contributors agreement.

nikolaykasyanov:
@swisspol thanks, will do.

nikolaykasyanov:
I've updated the commit, will review further uses of `-fileExistsAtPath:` a bit later.

nikolaykasyanov:
@swisspol I did some basic audit of `-fileExistsAtPath:` calls, you may find results in the checklist.

swisspol:
Thanks for the extensive audit 👍 

I think the right approach would be an extension method called `-fileExistsAtPath:followSymlink:` then replace **all** the call sites with this new method.

This will be a lot easier to maintain and audit than searching for 2 methods: the original `NSFileManager` one and then your new one.

Once you have this done, I can quickly do a search for `followSymlink:` in the diff on GitHub and review the value 😄

nikolaykasyanov:
@swisspol great idea! We'll also need `-fileExists:directory:followSymlink:`, probably.

nikolaykasyanov:
@swisspol I replaced `-fileExistsAtPath:` with `-fileExistsAtPath:followSymlink:` everywhere except tests.

nikolaykasyanov:
...and in tests as well.

GitUpKit/Core/GCFoundation.m
```diff
@@ -19,10 +19,19 @@
 
 #import "GCPrivate.h"
 
-#if !TARGET_OS_IPHONE
-
 @implementation NSFileManager (GCFoundation)
 
+- (BOOL)fileExistsAtPath:(NSString*)path followSymlink:(BOOL)followSymlink
+{
```
swisspol: Minor: `{` should be on same line as method

GitUpKit/Core/GCFoundation.m
```diff
@@ -19,10 +19,19 @@
...
+  if (followSymlink) {
+    return [self fileExistsAtPath:path];
+  }
+
+  return [self attributesOfItemAtPath:path error:NULL] != nil;
```
swisspol: I'd rather use a single API rather than 2. Looking at the docs, something like that would work:
```
if (followSymlink) {
  path = [path stringByResolvingSymlinksInPath];
}
return [self attributesOfItemAtPath:path error:NULL] != nil;
```

swisspol:
Good work! The diff looks much better now. So `followSymlink` is `NO` everywhere then?

GitUpKit/Core/GCFoundation.m
nikolaykasyanov: Fair enough.

nikolaykasyanov:
@swisspol I have to review everything once again.

swisspol:
Thanks much!

nikolaykasyanov:
@swisspol ready for final review.

How should I squash the commits?

swisspol:
Great, thanks!

I think a single commit should be enough now considering how simple the final change is. What do you think?

nikolaykasyanov:
@swisspol done.

GitUp/Application/AppDelegate.m
```diff
@@ -560,7 +560,7 @@ - (IBAction)newRepository:(id)sender {
   if ([savePanel runModal] == NSFileHandlingPanelOKButton) {
     NSString* path = savePanel.URL.path;
     NSError* error;
-    if (![[NSFileManager defaultManager] fileExistsAtPath:path] || [[NSFileManager defaultManager] moveItemAtPathToTrash:path error:&error]) {
+    if (![[NSFileManager defaultManager] fileExistsAtPath:path followSymlink:NO] || [[NSFileManager defaultManager] moveItemAtPathToTrash:path error:&error]) {
```
swisspol: So the part of the condition doesn't follow symlinks but what about the 2nd?

GitUp/Application/AppDelegate.m
```diff
@@ -591,7 +591,7 @@ - (IBAction)cloneRepository:(id)sender {
       if ([savePanel runModal] == NSFileHandlingPanelOKButton) {
         NSString* path = savePanel.URL.path;
         NSError* error;
-        if (![[NSFileManager defaultManager] fileExistsAtPath:path] || [[NSFileManager defaultManager] moveItemAtPathToTrash:path error:&error]) {
+        if (![[NSFileManager defaultManager] fileExistsAtPath:path followSymlink:NO] || [[NSFileManager defaultManager] moveItemAtPathToTrash:path error:&error]) {
```
swisspol: Same question

GitUpKit/Core/GCFoundation.m
swisspol: I don't see the change I suggested in there?

nikolaykasyanov: It seems like I lost the fix somehow while rebasing/squashing, sorry.

GitUp/Application/AppDelegate.m
nikolaykasyanov: `-[NSFileManager moveItemAtPathToTrash:error:]` uses `-[NSFileManager moveItemAtPath:toPath:error:]` in the end, and the latter doesn't move symlink destination, only link itself is moved:

>If the last component of srcPath is a symbolic link, only the link is moved to the new path; the item pointed to by the link remains at its current location.

It looks like 2nd part of the condition doesn't follow symlinks either.

swisspol: Per your quote, it does, but not the last one. So this is a different behavior than your new API which doesn't follow any symlink, right?

nikolaykasyanov: Fair enough. So we need `-moveItemAtPath...` which either follows all symlinks or follows no symlinks.

swisspol: Well let's sync this through: what is clear is that both APIs must have the same behavior with regard to symlinks. The question is then: what should be the correct behavior? There are two cases:
- symlink in the last path component 
- symlink anywhere else

I think symlink in the last path component is when you need an option to follow or not. For symlinks elsewhere, they should always be followed because the alternative doesn't make much sense.

swisspol:
Looks like -moveItemAtPathToTrash: already has the behavior we want: follows intermediary symlinks but not last one.

So the new API argument should be renamed "followLastSymlink" and the behavior modified as such. Then everything is consistent.

nikolaykasyanov:
@swisspol thanks, sounds good, I'm going to look into it.

nikolaykasyanov:
@swisspol ready for review, I've also added some tests for `-fileExistsAtPath:followLastSymlinks:`.

GitUpKit/Core/GCFoundation.h
```diff
@@ -15,10 +15,12 @@
 
 #import <Foundation/Foundation.h>
 
+@interface NSFileManager (GCFoundation)
+- (BOOL)fileExistsAtPath:(NSString*)path followLastSymlink:(BOOL)followSymlink;
```
swisspol: Arg name should be `followLastSymlink`

GitUpKit/Core/GCFoundation.m
```diff
@@ -19,10 +19,26 @@
 
 #import "GCPrivate.h"
 
-#if !TARGET_OS_IPHONE
-
 @implementation NSFileManager (GCFoundation)
 
+- (BOOL)fileExistsAtPath:(NSString*)path followLastSymlink:(BOOL)followSymlink {
+  NSDictionary *pathAttributes = [self attributesOfItemAtPath:path error:NULL];
```
swisspol: `*` next to var type

GitUpKit/Core/GCFoundation.m
```diff
@@ -19,10 +19,26 @@
...
+
+  if (followSymlink && [[pathAttributes fileType] isEqualToString:NSFileTypeSymbolicLink]) {
+    path = [self destinationOfSymbolicLinkAtPath:path error:NULL];
+
```
swisspol: Remove empty line

swisspol:
Good idea for the tests. As soon as you rebase and squash, I'll merge this in. Great work!

nikolaykasyanov:
@swisspol great!

I fixed code style issues in tests as well.

nikolaykasyanov:
@swisspol rebased & squashed.

nikolaykasyanov:
@swisspol thanks for your guidance!

Issue #102, opened 2016-03-31
Add option to not show Welcome screen on startup


sindresorhus:
I only open git repos from the command-line into GitUp and having the Welcome screen appear is kinda annoying when I would never use it.

sindresorhus:
@swisspol Would you prefer to have such option in `Preferences...` or directly in the Welcome screen UI? I've seen examples of both on OS X.

swisspol:
I think it should be in preferences?

larry-xmos:
+1

samchesney:
I'd appraise this as I do the same - start GitUp from the command-line, and then have to close the welcome screen that has opened on top of other windows on my other monitor.

sindresorhus:
In 1.0.5 it became considerably more annoying, as it's now shown every time GitUp is activated, not just on startup...

mtrudel:
Perhaps it belongs in preferences as well, but I think it's a reasonable default to not display it when invoked from the command line. Chances are overwhelming that you'll want to be interacting with the git folder implied by where the command was invoked, and so displaying the welcome screen isn't relevant.

PR #205, opened 2016-08-23
Add preference to disable welcome window (default NO)

mtrudel:
Per #102, this adds a preference to disable display of the welcome screen. It defaults to NO, so existing behaviour is preserved.

### DIFF
--- a/GitUp/Application/AppDelegate.m
+++ b/GitUp/Application/AppDelegate.m
@@ -107,6 +107,7 @@
     kUserDefaultsKey_FirstLaunch: @(YES),
     kUserDefaultsKey_DiffWhitespaceMode: @(kGCLiveRepositoryDiffWhitespaceMode_Normal),
     kUserDefaultsKey_EnableVisualEffects: @(NO),
+    kUserDefaultsKey_ShowWelcomeWindow: @(YES),
   };
   [[NSUserDefaults standardUserDefaults] registerDefaults:defaults];
 }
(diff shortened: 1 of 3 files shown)


AlexIzh:
@mtrudel suggestion: change title of settings to positive. For example: 
`Show welcome screen when startup` or `Start with welcome screen` or something like this. 
Reasons: 
1) Almost all settings have positive phrases in this app (I think all should be in one style)
2) Positive phrase is used more often and intuitively more understandable

mtrudel:
@AlexIzh I don't disagree - this was just the most expedient thing to get done. Happy to change; will defer to @swisspol for guidance.

swisspol:
Can you add a screenshot of the updated preference window?

mtrudel:
untitled

Upon mature reflection, I think I agree with @AlexIzh - the negative case is too confusing

swisspol:
Yes, just make it "Show welcome window". To be clear if this setting is off, then the window is never shown, right?

mtrudel:
Pushed an amended branch with positive logic

Per @swisspol's last, this displays 'Show welcome window' in preferences and defaults to YES, which preserves existing welcome screen behaviour. Disabling that checkbox in preferences will cause the welcome screen to not be shown under any circumstances.

swisspol:
All good then!

salutis:
@mtrudel Thanks!

PR #213, opened 2016-09-11
Fixed off-center search field placeholders

antons:
Off center placeholder is particularly bothersome on macOS Sierra. Even the default title was [drawn incorrectly](https://cloud.githubusercontent.com/assets/129705/18415307/357c6a88-77f6-11e6-869f-8d365617ac68.png) on launch. `[_searchField setNeedsLayout:YES]` forces search field to re-center the placeholder, both on launch, and during indexing.

Tested on macOS Sierra GM, and macOS El Capitan 10.11.5.

I removed your previous workaround for indexing (extra whitespace). Do you recall if you’ve tried `[_searchField setNeedsLayout:YES]` on Yosemite? If I recall correctly, search field centering had more bugs there, so this new workaround might not work. But I don’t have a machine with 10.10 to test this.

### DIFF
--- a/GitUp/Application/Document.m
+++ b/GitUp/Application/Document.m
@@ -22,6 +22,10 @@
 
 #import <GitUpKit/XLFacilityMacros.h>
 
+#ifndef kCFCoreFoundationVersionNumber10_12
+#define kCFCoreFoundationVersionNumber10_12 1348.1
+#endif
+
 #define kWindowModeString_Map @"map"
 #define kWindowModeString_Map_QuickView @"quickview"
 #define kWindowModeString_Map_Diff @"diff"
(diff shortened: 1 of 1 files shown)


swisspol:
The workarounds for this AppKit bug work on 10.8-10.11, so it's best not to change that code. Instead add a code path that's only for 10.12 and later.

antons:
@swisspol I think placeholders were not centered before 10.10. You’re right, it’s best to avoid a regression. I updated the fix to have two separate code paths. What do you think?

antons:
Ah, of course NSAppKitVersionNumber10_11 was declared only on 10.12 SDK, and CI runs with 10.11 SDK.

swisspol:
You'll want to use `kCFCoreFoundationVersionNumber` like in other places in the code.

Then add something like this at the top of the file:
```
#ifndef kCFCoreFoundationVersionNumber10_12
#define kCFCoreFoundationVersionNumber10_12 ?
#endif
```

For the actual value of `kCFCoreFoundationVersionNumber10_12`, you'll have to just print it `kCFCoreFoundationVersionNumber` in the debugger.

antons:
Thanks. Done.

swisspol:
Looks good. Did you use the format source script?

antons:
@swisspol Homebrew now installs clang-format 4.0.0. I modified format sources and ran it. It did not produce any changes.

swisspol:
Thanks, I'll wait until Sierra is officially out to merge this in.

PR #231, opened 2016-10-19
Fix crash if macOS version less than 10.12

vovkasm:
It is incorrect api usage in call to [NSString stringByPaddingToLength]: third argument "startingAtIndex" - index of padString, not original, so it should be always null in our case.


swisspol:
How do you reproduce this crash?

vovkasm:
Simple run app on macOS 10.11.6 with Xcode 7.3.1 and open any git repo.
I don't see repo window, but instead see this in Xcode console.
screenshot

vovkasm:
Ok, and of course: `I AGREE TO THE GITUP CONTRIBUTOR LICENSE AGREEMENT`

swisspol:
Does it crash if you use the official build? Because it doesn't crash for me on 10.11 and nobody else has reported this issue which would be extremely visible.

vovkasm:
It shouldn't because bug was added after v1.0.6 tag in 9c73a6463

swisspol:
Ah, thanks!

PR #265, opened 2017-01-10
Add a user preference for choosing a custom font in diff views. Fixes #87.

### LINKED ISSUE #87: Custom font selection for diffs
GitUp is a wonderful piece of software that makes my job easier every single day, so thanks!

The only thing that would make it more fantastic for me is if there was a way to change GitUp’s default typeface for viewing diffs to another monospace typeface of the user’s choice.

bosmacs:
As noted in enhancement request #87, the default font for diff views is quite small, and better yet, it would be nice to be able to choose the default font. This pull request implements that feature, and bumps up the default font size slightly to 11pt from 10pt. 

An entry in the preference window has been added for choosing the diff font:

screen shot 2017-01-09 at 8 43 02 pm

A corresponding NSUserDefaults key has been added to GitUpKit. GIDiffContentsViewController observes this key, and when it is changed, it forces GIDiffView to update the global font attributes, and recreates GIDiffView instances so the effect is seen immediately.

### DIFF
--- a/GitUp/Application/AppDelegate.m
+++ b/GitUp/Application/AppDelegate.m
@@ -37,6 +37,8 @@
 
 #define kWelcomeWindowCornerRadius 10
 
+#define kDefaultFontSize 11
+
 #define kInstallerName @"install.sh"
 #define kToolName @"gitup"
 #define kToolInstallPath @"/usr/local/bin/" kToolName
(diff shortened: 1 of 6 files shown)


bosmacs:
Thanks for pointing this out. I'll try to look into it when I have some time.

douglashill:
I’ve submitted my alternative PR as https://github.com/git-up/GitUp/pull/283. I thought it was good to put it up as well because the approach I’m taking is a bit different.

PR #283, opened 2017-02-25
Font scale

douglashill:

This branch adds a slider to let users change the font size of the diff and commit message views.

prefs

size changes

This addresses the font size requests in the comments of https://github.com/git-up/GitUp/issues/87 — but not changing the typeface itself which is what the original comment is about. See also the alternative PR https://github.com/git-up/GitUp/pull/265 while adds very similar functionality, including changing the typeface. I’m making this pull request because I think my branch has some advantages over other the other PR.

## Design goals

Users should be able to control how large text on computers because software authors can’t know in advance what size would be most comfortable for the user to read due to factors such as visual impairments, screen pixel density and the distance between the screen and the user.

This branch aims to lay a flexible foundation for allowing any part of GitUp to be scalable, and implements scalability in the diff and commit edit views, because I think these are the two views where readability is most important. Using the same scaling in other views is open as possible future work.

Letting the user specify a scale but the application specify the typefaces and relative sizes seems like the best trade-off between allowing the user to see the app comfortably without asking them to be the designer. Also different views have use different fonts and relative sizes. Therefore, the user does not set an explicit typeface name or numerical size and instead a slider without value labels has been added to General Preferences to set the scale. This (deliberately) results in the exact size used being hard to find.

The scale factor is stored in user defaults and this value happens to match the font size of the diff view. Users with unusual requirements for especially large (or small) text can set this to any value using the `defaults` tool.

## Implementation notes

One limitation of storing the font size in the standard user defaults is that this is global state so an app using GitUpKit that wants to show two diff views at the same time would not be able to make them use different font sizes. I don’t think this use case is particularly important.

As a framework, GitUpKit should not register default values in user defaults, and it can’t guarantee that the app using GitUpKit registers a sensible default for this key. Therefore I made GitUpKit always read the font size from user defaults wrapped in a `GIFontSize()` function which ensures the value is greater than zero.

I put the `GIFontSize` function in `GIAppKit.h/m` because that’s where other GitUpKit user defaults keys are defined. I didn’t think this fits in in `GIFunctions` because it’s only available in the Mac version of GitUpKit. Putting it in `GIFunctions` wrapped in `#if __GI_HAS_APPKIT__` would work too but makes less sense to me.

I thought it would be messy to do the `#ifdef __cplusplus` like in `GIFunctions` for this one function, so I used `FOUNDATION_EXPORT`. (My preferred way to do this would be to drop the `#ifdef __cplusplus` from `GIFunctions` and define a `GIEXPORT` used for every public function.)

Naming of the user defaults key: I mirrored `GICommitMessageViewUserDefaultKey_ShowMargins` but removed the `CommitMessageView` part because this one is more general.

I decided to store the most obvious thing in user defaults: the font size used in the diff view in points. This simplicity comes at the cost of making the slider in preferences more complicated as a linear slider would not be very good: the difference between 10 and 11 points is much more important than the difference between 18 and 19 points. A value transformer is used to map between the discrete slider value and a non-linear array of font sizes that match the preset sizes in the Mac system font picker.

### DIFF
--- a/GitUp/Application/AppDelegate.m
+++ b/GitUp/Application/AppDelegate.m
@@ -106,6 +106,7 @@
     GICommitMessageViewUserDefaultKey_ShowInvisibleCharacters : @(YES),
     GICommitMessageViewUserDefaultKey_ShowMargins : @(YES),
     GICommitMessageViewUserDefaultKey_EnableSpellChecking : @(YES),
+    GIUserDefaultKey_FontSize : @(GIDefaultFontSize),
     kUserDefaultsKey_ReleaseChannel : kReleaseChannel_Stable,
     kUserDefaultsKey_CheckInterval : @(15 * 60),
     kUserDefaultsKey_FirstLaunch : @(YES),
--- a/GitUp/Application/FontSizeTransformer.h
+++ b/GitUp/Application/FontSizeTransformer.h
@@ -1,0 +1,21 @@
+//  Copyright (C) 2015-2017 Pierre-Olivier Latour <info@pol-online.net>
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#import <Foundation/Foundation.h>
+
+/// Transforms between the linear slider in Preferences to a non-linear list of font sizes matching the system font picker.
+@interface FontSizeTransformer : NSValueTransformer
+
+@end
--- a/GitUp/Application/FontSizeTransformer.m
+++ b/GitUp/Application/FontSizeTransformer.m
@@ -1,0 +1,55 @@
+//  Copyright (C) 2015-2017 Pierre-Olivier Latour <info@pol-online.net>
+//
+//  This program is free software: you can redistribute it and/or modify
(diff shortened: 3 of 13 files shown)


outpunk:
@swisspol Pleeeeeease!

fab1an:
This is the only reason I don't use GitUp.

douglashill:
@outpunk @fab1an @JanX2 @reddavis @azizhk @sindresorhus @egnha @salutis

You can always build GitUp yourself from this branch, but to make things a bit easier I’ve published a release build from my fork at https://github.com/douglashill/GitUp/releases

egnha:
@douglashill Thanks very much for making the build available. Works great!

JanX2:
@douglashill That’s great. Thank you!

tulios:
What is missing in this PR?

Pleasurazy:
This is the only pain I using GitUp.

egnha:
@swisspol First of all, thanks very much for GitUp. It's one of the reasons I'm still on a Mac.

With the closing of this PR, am I to understand that a user option to configure font size won't be forthcoming? Or are you hoping to implement it differently from this PR?

swisspol:
It's a very large diff that's been around for a while. The problem is that I don't have bandwidth to properly review / merge / maintain such a change, sorry.

douglashill:
Thanks for clearing this up @swisspol. I agree better closed than left around.

@egnha or anyone else who wants this feature: I encourage you to build GitUp yourself or use [the build I've made available](https://github.com/douglashill/GitUp/releases) with these changes. It's really awesome that GitUp is open source so we can do this. Thank you @swisspol! ❤️

lourenci:
I think this could be opened with the new maintainer @lucasderraugh.

lolgear:
@lucasderraugh Could you open it? I have a problem with font size. It is so small :(

lucasderraugh:
I don’t have the ability to because the branch has since been deleted. If @douglashill wants to reopen we can definitely look at it again.

douglashill:
This is on my list to look into how viable it is to bring my text branch onto current master. Will hopefully get to this at a less busy time.

lucasderraugh:
@douglashill If you ever do want to merge, can we use this #615 as our new issue. I'd like an open issue on this.

PR #301, opened 2017-04-07
Correct working directory for git-lfs

### LINKED ISSUE #292: Crash when using Git LFS
I have a repository using GitHub Enterprise and Git LFS, and since updating to 1.0.7, I get frequent crashes when working with this repo. Strangely, the OS crash dialog does not appear after these crashes, so I'm not sure how to collect a crash log.

Steps to reproduce:
1. Open a repo that uses Git LFS
2. Modify one of the files tracked by Git LFS
3. Switch to the "Commit" view

Expected results: Should see the changes in the file from step 2
Actual results: GitUp crashes

I've also seen crashes in the "Map" view when pulling latest changes from a remote where the changes being pulled include an update to a file tracked by LFS.

joshhinman:
The Git LFS filter was using the `.git` directory instead of the working directory. This change fixes #292


swisspol:
Oh nice!

Issue #144, opened 2016-05-12
Failed writing snapshots on non-HFS volumes


sjackman:
I frequently get the error message "Failed writing shapshots"
How can I troubleshoot this error?

swisspol:
Only thing I can think of is that you're out of disk space? There might be more info in the console log.

Worst case, build and run app from Xcode and put a breakpoint on this error.

ccodewa:
same problem for me on 1.05 (1017).
Happens only on git repos mounted on NFS.
1.0.4 works fine.

swisspol:
I'm guessing that's because non-HFS volumes don't support `exchangedata()`: d78d6d3868e8b725238b13a56522bf387f6f4a67

swisspol:
Note that using GitUp on repos which are not on HFS volumes is not supported anyway. File monitoring might not work correctly anyway.

ccodewa:
Thank you Pierre.
What about disable the browse snapshot button and menu item on non-HFS filesystems repos instead of the "Failed writing shapshots" popup error?

sjackman:
Even if it's not officially supported, `GitUp` has been working just fine for me up until now on a NFS file system.

stigkj:
I get the same error on OpenZFS. The com.apple.mimic_hfs flag is set, but does not help. Just in case anyone is wondering :-)

neeklamy:
I’ve just received this error on making a commit.

The repository is on an ext4 filesystem, and I’m connecting over SSHFS.

Could [fcntl](https://www.freebsd.org/cgi/man.cgi?query=fcntl) F_FULLFSYNC be used? I found this nugget of information:

> fsync may be really slow, but if it is not reliable that is a defect in the filesystem, block layer, or disk drive. On Mac OS X, fsync is documented not to be reliable, but there is at least a way around it using an F_FULLSYNC fcntl. On Linux (relatively recent versions anyway), fsync is documented to be reliable, i.e. to accomplish the purpose one might have for calling it in the first place.
> – butlerm ([direct](https://lwn.net/Articles/668353/)), from [Files are Hard](https://lwn.net/Articles/667788/)

Maybe this would be helpful too? [Everything You Always Wanted to Know About Fsync()](http://blog.httrack.com/blog/2013/11/15/everything-you-always-wanted-to-know-about-fsync/).

szhu:
[...]

PR #320, opened 2017-06-07
Check if volume supports exchangedata before use

antons:
This fixes incompatibility with APFS. I’m not sure if remove() and rename() is the right way to replace exchangedata(). Let me know if it’s not.

### DIFF
--- a/GitUpKit/Core/GCLiveRepository.m
+++ b/GitUpKit/Core/GCLiveRepository.m
@@ -19,6 +19,7 @@
 
 #import <sys/stat.h>
 #import <sys/attr.h>
+#import <sys/mount.h>
 
 #import "GCPrivate.h"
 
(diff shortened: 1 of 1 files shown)


salutis:
Consider adding `Close #144` if applicable.

zwaldowski:
On the appropriate OS, it might be appropriate to use `renamex_np` instead? It seems to accomplish something similar.

Also, for the fallback case, Darwin's [`rename`](http://www.manpagez.com/man/2/renamex_np/) is documented to remove first.

swisspol:
Thanks for sending this in.

I'd rather see a new utility C function that swaps 2 files. It would use `exchangedata()` if available and falls back to something else otherwise. Then update the call sites to use this new function.

PS: Note that `exchangedata()` is used in 2 places by the way.

PR #324, opened 2017-06-12
Avoid using exchangedata entirely

### LINKED ISSUE #144: Failed writing snapshots on non-HFS volumes

zwaldowski:
Alternative to #320. Would also close #144.

Fully atomic filesystem operations aren't worth the overhead in added complexity in checking for filesystem capabilities, now with APFS superseding atomic content replacement with atomic renaming; using a temporary file and moving into place practically works around any conflicts that may arise.

As far as using the non-streaming `NSKeyedArchiver`, I've opened a few large repos of mine and I haven't noticed any problems. I could be talked into using `NSFileManager` to do the move instead.

### DIFF
--- a/GitUpKit/Core/GCIndex.m
+++ b/GitUpKit/Core/GCIndex.m
@@ -581,8 +581,8 @@ cleanup:
   copyfile_state_free(state);
   CHECK_POSIX_FUNCTION_CALL(goto cleanup, status, == 0);
 
-  // Swap temporary copy and original file
-  CALL_POSIX_FUNCTION_GOTO(cleanup, exchangedata, fullPath, tempPath, FSOPT_NOFOLLOW);
+  // Move temporary copy.
+  CALL_POSIX_FUNCTION_GOTO(cleanup, rename, tempPath, fullPath);
   CALL_POSIX_FUNCTION_GOTO(cleanup, utimes, fullPath, NULL);  // Touch file to make sure any cached information in the index gets invalidated
 
   success = YES;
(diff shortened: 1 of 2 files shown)


swisspol:
Thanks for sending this in.

I'd rather see a new utility C function that swaps 2 files. It would use `exchangedata()` if available and falls back to something else otherwise. Then update the call sites to use this new function.

zwaldowski:
If `rename` is good enough for the case where atomic file content swapping is not available, why is it not good enough in the general case? POSIX itself guarantees `rename` is atomic. The implementation behind `NSDataWritingAtomic` only uses `atomic`. What atomicity case is not being addressed from using a side file in the first place?

Not only does `exchangedata` stop working under APFS, but it's been deprecated in the 10.13 SDK and will presumably disappear in future SDKs. The header recommends `renamex_np` as its replacement, which is also non-standard. A common "safe" swap function (which is not all that safe; the limitations of atomic file swapping have been documented) that covers all three code paths — including two unrelated syscalls just to determine filesystem capability — does not seem worth the added complexity.

I ask that you reconsider. Can you help me understand what context I am missing?

swisspol:
This is not about only atomicity, it's about swapping the contents of the files while preserving their metadata - on HFS+ this really matters. See https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man2/exchangedata.2.html.

PR #325, opened 2017-06-12
Modernize project for Xcode 9 warnings

zwaldowski:
Modernizes the project and fixes any warnings.

In particular, there's a minor renamefest and API change around `NSError` inouts, as described in [WWDC 2017 session 411, "What's New in LLVM"](https://developer.apple.com/videos/play/wwdc2017/411/). This could've been a source of use-after-free in production.


swisspol:
Thanks for sending this in but Xcode 9 is pre-release so let's hold on such changes until it's final.

zwaldowski:
I'll gladly rebase out any project changes, but the flaws addressed here exist whether or not there's a warning given for them. Allow me to enumerate the changes included:

- Implicit pointer-to-integer conversions. Implicit conversions are not guaranteed to have correct ARC lifetime semantics without additional annotation, here a bridging conversion.
- `()` function prototypes. By legacy C rules, function parameter lists declared as `()` can take any number of parameters of unknown types. They may be called incorrectly, which can leak or prematurely deallocate ARC pointers depending on the calling convention. The parameter lists of blocks used as r-values to a block l-value with an incomplete prototypes cannot be checked by the compiler.
- Implicit integer/enum conversion. `git_error_code` is not necessarily of type `int` and may lead to truncation.
- Capturing autoreleased `inout` objects in blocks. These have never been safe. Assigning to the captured pointer will `autorelease` the assigned object. The autorelease pool may pop in the during the execution of the block (for a variety of reasons; `-[NSDictionary enumerateKeysAndObjectsUsingBlock:]` is known to push its own pool), typically leading to a harmless but spurious `nil` error but potentially leading to a use-after-free in a multithreaded context.

Let me know if there's anything I can do to reduce the impact of the changes, but I feel like they're all legitimate issues. All of these warnings come from Clang open source and are already in production, so it's not merely Xcode beta weirdness.

PR #329, opened 2017-06-17
Check if volume supports exchangedata before use

antons:
This fixes incompatibility with APFS. As suggested in #320, I extracted a C function. If you’d like to see anything else changed, let me know.

I AGREE TO THE GITUP CONTRIBUTOR LICENSE AGREEMENT.

PR #351, opened 2017-08-18
Check if volume supports exchangedata before use

antons:
This fixes incompatibility with APFS.

After a big more digging I found that Apple added `renamex_np` which accepts a `RENAME_SWAP` option. It is the recommended way to implement atomic saves going forward.

For volumes and system combinations that support neither function the `GCExchangeFileData` will fail, as it did anyway. It doesn’t seem right to name `GCExchangeFileData` function to remove the `path1`, as noted in #329.

### DIFF
--- a/GitUpKit/Core/GCFunctions.h
+++ b/GitUpKit/Core/GCFunctions.h
@@ -39,6 +39,8 @@ NSString* GCFileSystemPathFromGitPath(const char* string);
 NSURL* GCURLFromGitURL(NSString* url);
 NSString* GCGitURLFromURL(NSURL* url);
 
+int GCExchangeFileData(const char* path1, const char* path2);
+
 void GCArrayApplyBlock(CFArrayRef array, void (^block)(const void* value));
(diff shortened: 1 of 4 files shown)


swisspol:
Looks good to me, thanks!

rgrove:
Is there anything preventing this from being merged? macOS High Sierra [will be released on September 25th](https://www.apple.com/macos/high-sierra/), and APFS isn't optional for SSD users.

bjtitus:
FWIW, I've been using this fork for the last several weeks without issue on the High Sierra betas and GM.

rgrove:
@swisspol Thanks for merging this.

macOS High Sierra was released today, but the latest stable build of GitUp still doesn't include this change. Anyone upgrading to High Sierra and APFS will encounter frequent "Failed writing snapshots" errors in GitUp. If it's not too much trouble, could you please release a new stable build?

iest:
Top work @antons! This has been bugging me for weeks (pun intended).

PR #352, opened 2017-08-18
Fixed snapshots on APFS

fjolnir:
This changes swapping to use `renamex_np` with `RENAME_SWAP` if available, otherwise falling back to `exchangedata`. I don't have a pre-sierra, nor one with an HFS+ volume to test on. So this has only been tested using High Sierra & APFS.


Kumuluzz:
@fjolnir I'm experiencing the exact same issue as you and it's really annoying. Could you go ahead and accept the CLA so I don't have to deal with this annoying bug anymore? 😄 
https://github.com/git-up/GitUp/blob/master/CONTRIBUTING.md

swisspol:
This has been fixed in #351

PR #363, opened 2017-10-14
Add namespace to methods in categories on Apple classes

fcanas:
In my experience it is a safer practice to namespace category methods on classes you don't own, especially in frameworks and libraries.


swisspol:
Yes this safer in theory you are correct. However this is not a _requirement_ for things to work as long you don’t use too commun method names. GitUp is fine as is but thanks for sending this in.

PR #384, opened 2017-11-30
make tag and branch name text fields single-line

DivineDominion:
GitUp taught me to hit ⌥+Enter to commit -- but in tag and branch name text fields, it inserted a line break. Now these text fields are in "Single Line Mode" as expected.

Would be even cooler if ⌥+Enter triggered the success-action instead of Enter, but at least it doesn't insert invalid line breaks anymore.


swisspol:
See #410

PR #388, opened 2017-12-04
By default, do not show invisible characters in the commit message area.

beatrupp:
I think the special characters in the message area of commits are irritating and should at least be disabled by default, so that adventurous users can enable them :)


swisspol:
This is like this by design, and people don’t mind, so let’s not change it, sorry.

beatrupp:
Ok not a problem. Still, this is the first app I have ever seen this default behaviour. Good thing is that there's an option to disable it  😅

Issue #397, opened 2017-12-20
Document window allows tabs to be opened, but tabs aren't actually usable


davidfmiller:
(After opening a new tab, clicking the "File → Open" menu item opens a new document window entirely.)

![screen shot 2017-12-20 at 1 11 05 pm](https://user-images.githubusercontent.com/12614/34228913-43f86fb2-e587-11e7-906d-4714eac4a466.png)

Seems like the quick solution is to remove the "Show Tab Bar" and "Show All Tabs" menu items from the "View" menu.

swisspol:
Tabs are not supported in GitUp. These items should not appear in "View" menu indeed.

davidfmiller:
I'd love to lend a hand on this one. I've forked the GitUp repo and am trying to get it to build, to no avail. I've looked through the source and have got things to the point where the linking fails on `libssh2.a`, `libsqlite3.a`, and `libcrypto.a`.

Trying to find a guide to get things going but haven't come across one in the project wiki or the repo itself... can anyone help this noob out? Thanks.

PR #399, opened 2017-12-21
Fix new tab behaviour by implementing newWindowForTab

lapfelix:
Fix for https://github.com/git-up/GitUp/issues/397

I use tabs in GitUp all the time so I disagree with disabling tabs altogether to fix the odd new tab button behaviour.

To fix the issue I simply implemented the `newWindowForTab:` method in NSDocumentController and linked it to the "newRepository" method in the app delegate.

A future improvement might be to show the welcome window instead to open a recent repository rather than create a new one ?


salutis:
@lapfelix 

> I use tabs in GitUp all the time so I disagree with disabling tabs [...]

+1 I use tabs all the time too.

swisspol:
I'll check-in a different fix.

PR #402, opened 2017-12-29
Minor changes in tips

pravdomil:
[diff](402/files)


swisspol:
Thanks for sending this in but I’d rather keep the current convention.

pravdomil:
@swisspol well across all macOS interfaces Apple the convention is use ⌘ for Command

https://developer.apple.com/macos/human-interface-guidelines/menus/menu-anatomy/

salutis:
+1 for ⌘

PR #409, opened 2018-01-23
Mode control width fix

m-konovalov:
Hey! 

Just mode control width fix. 
Before:
screen shot 2018-01-23 at 09 08 12
After:
screen shot 2018-01-23 at 09 07 58


swisspol:
Thanks

Issue #76, opened 2016-03-12
browse a file's commit history


RBornat:
I'd like to be able to browse a file's commit history. Information about commits is available via git log, but it's hard to find the commits in a very long list (mine is over 20 years long). 

Or am I missing something obvious?

swisspol:
No, there is no such feature today.

mik01aj:
You can get close to it by using the search box. (Maybe just the search box needs improvement?)

kpowick:
@mik01aj As far as I can tell, the git-up search box only searches commit messages. Which, unless you included the file name in all your comments, will be unlikely to retrieve the full commit history (git log) of an individual file.

I agree that this is a valuable feature that should be added to git-up.  It's one of the only things keeping me from using git-up more often.

mik01aj:
@kpowick it can:

![image](https://cloud.githubusercontent.com/assets/3027415/13840666/1f711094-ec24-11e5-90af-f05d86d218d7.png)

![image](https://cloud.githubusercontent.com/assets/3027415/13840659/0fc0b7da-ec24-11e5-9339-6c7b2c322fa3.png)

kpowick:
Hi @mik01aj

I guess we're looking for different things. 

What I'm looking for is a way to see the complete commit history (git log) of a single, specific file (e.g. myprog.c).  Other products such as _SmartGit_, _GitKraken_, and _SourceTree_ have such a feature.  For _GitUp_, enabling _"Search commit diffs"_ does not provide similar functionality.

mik01aj:
@kpowick I meant yes, it's not the same, but it does get close. For me it's practical enough.

For the precise case, yes, sometimes I need the full history of a specific file, and then I use IntelliJ integration. It can even show git history for some specific part of the file. It'd be cool to have it in GitUp too but it's not a deal-breaker for me.

nathanpalmer:
Search by file would be huge. It's a big missing piece from moving over from Sourcetree for me.

Today I can search for a file

![screen shot 2016-12-13 at 3 41 40 pm](https://cloud.githubusercontent.com/assets/125599/21158311/cbc61636-c14a-11e6-83d2-36db923b8c47.png)

and then review the log for it

![screen shot 2016-12-13 at 3 42 00 pm](https://cloud.githubusercontent.com/assets/125599/21158317/d454d26a-c14a-11e6-8bbf-56cf0e7f8e79.png)

[...]

PR #423, opened 2018-02-24
Allow searching single file history with prefix "/" 

soleo:
## What is this PR trying to solve

Refer to #76 

This PR enables a user to search a single file's commit history. The equivalent git command would be `git log --follow path/to/filename`. The difference between the git command and a gitUp single file search is we're using prefix `/` to indicate that we only want the file history.

e.g.

If you want to search any changes for file `package.json`, you'd use `/package.json` in search.

If you want to search for the keyword `package.json` in commit messages, you'd use `package.json` instead.

## What it looks like
[
screen shot 2018-02-25 at 1 54 22 pm
](https://user-images.githubusercontent.com/635858/36645962-51ec8674-1a36-11e8-95a1-50a6f0c2e7b6.png)

## Known issue

When searching in a large repo, it is slow

## Contribution Agreement

### DIFF
--- a/GitUpKit/Core/GCLiveRepository.m
+++ b/GitUpKit/Core/GCLiveRepository.m
@@ -443,7 +443,6 @@ static void _StreamCallback(ConstFSEventStreamRef streamRef, void* clientCallBac
   NSString* path = [self.privateAppDirectoryPath stringByAppendingPathComponent:kCommitDatabaseFileName];
   _updatingDatabase = YES;
   dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_LOW, 0), ^{
-
     NSError* error;
     GCRepository* repository = [[GCRepository alloc] initWithExistingLocalRepository:self.repositoryPath error:&error];  // We cannot use self because we access the repo on a background thread
     GCCommitDatabase* database = repository ? [[GCCommitDatabase alloc] initWithRepository:repository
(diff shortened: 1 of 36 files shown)


GitUpKit/Core/GCLiveRepository.m
```diff
@@ -1008,8 +1008,18 @@ static BOOL _MatchReference(NSString* match, NSString* name) {
 - (NSArray*)findCommitsMatching:(NSString*)match {
   XLOG_DEBUG_CHECK(_database);
   NSMutableArray* results = [[NSMutableArray alloc] init];
+  
   match = [match stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
-  if (match.length >= kMinSearchLength) {
+  bool searchFileHistoryOnly = [match hasPrefix:@"/"];
+  // Search file history directly
+  if(match.length >= kMinSearchLength && searchFileHistoryOnly) {
```
swisspol: This needs to be (kMinSearchLength + 1) because of prefix /

soleo: Will change it in next commit.

soleo:
@swisspol Fix the issues you mentioned in the comments. However, I cannot lint the code with `format-source.sh` because homebrew now uses clang-format v7.0.0 instead of v6.0.0. Should we update the version and push another commit with v7.0.0-linted code or shall we use do another PR to fix the version?

swisspol:
Try rebasing on master and running `format-source.sh` again (only pick changes on your files though).

PR #458, opened 2018-06-06
Assign 2x app icons for retina screens

sfsam:
The AppIcon asset already has the correct images, but some of the 2x slots (16, 128, 256) were unassigned. This led to slightly pixelated app icons on retina screens. This pull request simply assigns the 2x icons using the existing png assets.


swisspol:
Thanks!

PR #460, opened 2018-06-08
Fixes interaction with toolbar mode picker on Mojave

zwaldowski:
Mojave moves several controls to paint using subviews instead of cells, causing the double-click hit test checking to fail. This addresses that, and allows building on the 10.14 SDK unmodified.


swisspol:
Looks great, thanks!

PR #496, opened 2019-01-21
Improving informational overlay so it does occlude UI

nathanborror:


Gatada:
I'm just jumping in here, so this may not be relevant - but if these overlays show text - a better approach would be a dynamic delay depending on word and/or character count.

For 10 or more words I find 0.3 seconds per word to work well, and for less than 10 words 0.2 seconds. This is based on personal experience and information I read online about reading speed:

- an adult reads about 250 words per minute on an average.
- a slow reader usually reads about 150 to 200 words per minute.
- for technical contents, the average reading speed is about 50-75 words per minute.

lucasderraugh:
@nathanborror I've recently taken over as temporary maintainer of GitUp. If I can get an overview of visual changes here and squash changes into a single commit, I'd love to start the discussion on merging this in.

lucasderraugh:
I'm going to agree with @Gatada on this one. Unfortunately with longer sentences, 1.25 seconds simply isn't enough time to read the following image text. I agree that it's an improvement for the smaller branch switching text, but it become unusable for longer text. To accept this change, we'd want something dynamic so that all text could be read, but that it also is shorter than the current timeframe. I'll keep this open for a week or 2, but I'll close it if there isn't new activity. If someone wants to take the word counting approach, I would approve of that type of change.
Screen Shot 2019-06-25 at 11 34 34 PM

lucasderraugh:
As it seems nobody is probably actively working on this feature, I'm going to close this PR for cleanup purposes. If anyone would like to augment this change and reopen the PR I'd be happy to take it in.

PR #532, opened 2019-06-23
Dark mode support

zwaldowski:
Completes #463.

### TODO

- [x] Investigate `controlBackgroundColor`.

---

This PR brings the dark mode implementation from @saagarjha and @douglashill into upstream.

In addition to their (awesome) work, this PR:

- Better supports the color scheme changing at runtime
- Makes some icon buttons light up instead of darkening when pressed in Dark Mode
- Makes rows in the configuration editor are readable in Dark Mode
- Automatically supports the main graph/map view desktop-color tinting in Dark Mode
- Draws the separator grid lines in diff views without a "hatched" effect
## Review Notes

Due to parts of the app being unusable when dark mode isn’t supported in that screen, this PR isn’t as small and granular as would be ideal. However, the diff was specifically adjusted to limit unnecessary churn, such as avoiding the automatic rewriting of xibs. I failed to identify good points to break up what remains; if you find chunks you’re rather me split out to review separately, I’d be happy to do so (through the magic of GitUp!).

## Screenshots

<details>
<summary>10.15 (Dark)</summary>
<a target="_blank" rel="noopener noreferrer href="https://user-images.githubusercontent.com/170812/60307735-c892c100-9913-11e9-82f1-e19e67bfdbbc.jpg">10.15 Dark #1</a>
<a target="_blank" rel="noopener noreferrer href="https://user-images.githubusercontent.com/170812/60307736-c892c100-9913-11e9-961d-37aa7740178d.jpg">10.15 Dark #2</a>
<a target="_blank" rel="noopener noreferrer href="https://user-images.githubusercontent.com/170812/60307737-c892c100-9913-11e9-8534-02f0b8db1848.jpg">10.15 Dark #3</a>
<a target="_blank" rel="noopener noreferrer href="https://user-images.githubusercontent.com/170812/60307738-c892c100-9913-11e9-902d-8b2a4cb41ad0.jpg">10.15 Dark #4</a>
<a target="_blank" rel="noopener noreferrer href="https://user-images.githubusercontent.com/170812/60307739-c892c100-9913-11e9-8864-0d35d701e985.jpg">10.15 Dark #5</a>
</details>
<details>
<summary>10.15 (Light)</summary>
<a target="_blank" rel="noopener noreferrer href="https://user-images.githubusercontent.com/170812/60307741-c892c100-9913-11e9-9cea-09783c951437.jpg">10.15 Light #1</a>
[...]

PR #533, opened 2019-06-26
Allow files in diff views to be dragged outside of the app

zwaldowski:
Modernizes the table drag/drop in diff file views so it can pass file names, URLs, and promises to drags that finish outside the app.

I often find myself in the situation, particularly when resolving conflicts, that I want to open a file in an editor that would not be opened by clicking “Open with Default Editor”. This supports that use case and many more.

| Dragging a file or files from… | …to… | …will… |
|--------------------------------|:----------------------------------:|-------------------------------------:|
| any diff file list | another file list in the same view | stage/unstage/split/etc. |
| any diff file list | any other file list | do nothing. |
| the working copy | an icon in the macOS Dock | try to open them in the app. |
| the history | an icon in the macOS Dock | open a copy of the blobs in the app. |
| any diff file list | Terminal | paste relative paths to the files. |
| and diff file list | Terminal, with ⌥ held | paste absolute paths to the files. |

Screen Recording 2019-06-25 at 12 18 58 PM

⚠️ Note, the screen recording shows Dark Mode but this PR does not include #532, it was just recorded from a combined branch.

### DIFF
--- a/GitUpKit/Components/GIDiffFilesViewController.m
+++ b/GitUpKit/Components/GIDiffFilesViewController.m
@@ -22,7 +22,7 @@
 #import "GIInterface.h"
 #import "XLFacilityMacros.h"
 
-#define kPasteboardType @"GIDiffDelta"  // Raw unretained pointer which is OK since pasteboard is use within process only
+static const NSPasteboardType GIPasteboardTypeFileRowIndex = @"co.gitup.mac.file-row-index";
 
 @interface GIFileCellView : NSTableCellView
 @end
@@ -40,10 +40,7 @@
 @implementation GIFileCellView
 @end
 
-// Override all dragging methods to ensure original behavior of NSTableView is gone
-@implementation GIFilesTableView {
-  NSDragOperation _dragOperation;
-}
+@implementation GIFilesTableView
 
 - (BOOL)becomeFirstResponder {
   if (![super becomeFirstResponder]) {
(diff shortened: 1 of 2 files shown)


GitUpKit/Components/GIDiffFilesViewController.m
```diff
@@ -242,6 +196,10 @@ - (void)setSelectedDeltas:(NSArray*)deltas {
   [_tableView scrollRowToVisible:indexes.firstIndex];
 }
 
+- (GIDiffFileData*)objectAtIndexedSubscript:(NSInteger)index {
```
lucasderraugh: Eh, this is really weird to have on a non indexable data-type class. Can we just add generics to the mutable array instead?

zwaldowski: > so this can be accessed outside this class

It's not even outside the file, it's exclusively from one `GIDiffFilesViewController` to another, it was just kvetching about accessing the ivar.

I'll revisit it, thanks!

zwaldowski:
Updated.

GitUpKit/Components/GIDiffFilesViewController.m
zwaldowski: Done!

lucasderraugh:
This all seems like an improvement to me. The only thing I notice is that dragging to Terminal from either the working directory, the index, or history all seems to result in a relative path from the directory level, never an absolute path as your chart says. Looks great other than that.

zwaldowski:
Good call-out. That might be fine (since it's not a regression), the primary intent here is definitely other apps, but I'll double-check it.

zwaldowski:
OK! This was my typo in the table. Terminal does no special adjudication for the pasteboard; it always requests strings — thus choosing the relative path — unless Opt is held during the drag. This is fine for now. I think we could get away with sending just the file URL all the time, but in the spirit of not having any regressions I'm leaving the PR as it is. I've updated the table in the PR description accordingly.

Issue #463, opened 2018-06-12
Support for macOS Mojave's dark mode


saagarjha:
I thought I'd create an issue to track dark mode support on macOS Mojave. [I have a branch set up](https://github.com/saagarjha/GitUp/tree/mojave) that has some preliminary support (it also has what I think is a slightly nicer fix for #459, and some other improvements). Here's what it looks like currently:

<img width="890" alt="screen shot 2018-06-12 at 00 37 40" src="https://user-images.githubusercontent.com/13786931/41276345-d2cbf180-6dd8-11e8-8c3b-024b71e1c5f4.png">
<img width="890" alt="screen shot 2018-06-12 at 00 38 31" src="https://user-images.githubusercontent.com/13786931/41276386-f0c131f0-6dd8-11e8-9dd1-918c8fba8169.png">
<img width="366" alt="screen shot 2018-06-12 at 00 38 50" src="https://user-images.githubusercontent.com/13786931/41276402-fbd72856-6dd8-11e8-8d46-150ed23011f7.png">

Is there anyone else working on this?

sdgandhi:
Love it. Small UI feedback: full white is harsh on dark backgrounds, try 80-90% opacity white. And the other key colors (red, green, etc.) probably need to be dulled a little too.

saagarjha:
I'm sorry, what white are you talking about? The only white I see is the app icon and the text. As for the colors, these have already been tweaked a bit to work a bit better on dark mode, but I've been to lazy to fine-tune them more. I'm not particularly great on chromatic design, so most of my stuff is just "this color looks about right", so if you have suggestions for colors I'd be happy to look into using them.

saagarjha:
I'm using the standard system `NSColor.textColor`, which appears to be black in light mode and white in dark mode. This appears to be what most apps are doing as well.

Regarding the opacity of the the other colors, I'm not quite sure what you mean. All the colors are 100% opaque, but I've toned down and darkened the colors for dark mode:

<img width="507" alt="screen shot 2018-09-27 at 12 33 43" src="https://user-images.githubusercontent.com/13786931/46170049-94d61e00-c251-11e8-9c3b-b90aa2a97575.png">
<img width="507" alt="screen shot 2018-09-27 at 12 33 59" src="https://user-images.githubusercontent.com/13786931/46170063-9d2e5900-c251-11e8-9429-9c87142a9d21.png">

Is this what you were asking?

sdgandhi:
Yeah that looks good

[...]

PR #534, opened 2019-06-27
Bump deployment target to 10.10

zwaldowski:
As discussed in #463 and #532.

Not much to comment on here, should be no functional change except (extremely) greasing the wheels for #532. It takes the shortest path to get to a warning-free build on the newer deployment target.


GitUp/Application/Document.m
```diff
@@ -1838,12 +1826,13 @@ - (IBAction)editConfiguration:(id)sender {
 
 - (void)_setSearchFieldPlaceholder:(NSString*)placeholder {
   // 10.11 and earlier: search placeholders have the same length to work around incorrect centering.
-  if (kCFCoreFoundationVersionNumber < kCFCoreFoundationVersionNumber10_12) {
+  if (@available(macOS 10.12, *)) {
+  } else {
```
lucasderraugh: Maybe if `!@available`?

GitUp/GitUp.xcodeproj/project.pbxproj
```diff
@@ -485,7 +486,7 @@
 		E21DCAEC1B253847006424E8 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				GCC_DYNAMIC_NO_PIC = YES;
+				CODE_SIGN_IDENTITY = "Mac Developer";
```
lucasderraugh: Does this try and default to your current profile?

GitUpKit/Components/GISnapshotListViewController.m
```diff
@@ -62,7 +62,7 @@ - (instancetype)initWithRepository:(GCLiveRepository*)repository {
   return self;
 }
 
-- (void)viewWillShow {
+- (void)viewWillAppear {
```
lucasderraugh: All these NSViewController calls should be calling super, no?

Examples/GitY/GitY.xcodeproj/project.pbxproj
```diff
@@ -274,20 +274,32 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				ALWAYS_SEARCH_USER_PATHS = NO;
+				CLANG_ANALYZER_LOCALIZABILITY_NONLOCALIZED = YES;
```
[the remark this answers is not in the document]
zwaldowski: These changes are in the Examples projects, which are not governed by the project config files. I have no issue with making them that way.

GitUp/GitUp.xcodeproj/project.pbxproj
zwaldowski: Yes. The main app is set the same way, this is for the command-line tool.

GitUp/Application/Document.m
zwaldowski: Sadly, I’m like 99% sure it won’t work having tried it before. But once I get back to my desk, I’ll double-check. :)

GitUpKit/Components/GISnapshotListViewController.m
zwaldowski: Strictly speaking, no. All of the default implementations are no-ops and none of them are marked with `NS_REQUIRES_SUPER` to emit a warning. I have no issue with adding them for safety’s sake, though.

zwaldowski:
Will update to address comments shortly. Thanks!

GitUp/Application/Document.m
zwaldowski: Yeah, it does not. `@available does not guard availability here; use if (@available) instead`

GitUpKit/Components/GISnapshotListViewController.m
zwaldowski: Addressed in next push.

Examples/GitY/GitY.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
zwaldowski: Updated - all targets now use the shared config.

zwaldowski:
Updated. Making all the projects share the same config made for a nice cleanup, so this is a win all around.

GitUpKit/Components/GISnapshotListViewController.m
lucasderraugh: Huh, for some reason I thought it had `NS_REQUIRES_SUPER` on those calls, but I guess you're right. Anyways 😄...

Examples/GitY/GitY.xcodeproj/project.pbxproj
```diff
@@ -272,70 +287,15 @@
...
-				GCC_C_LANGUAGE_STANDARD = gnu99;
-				GCC_NO_COMMON_BLOCKS = YES;
-				GCC_WARN_64_TO_32_BIT_CONVERSION = YES;
-				GCC_WARN_ABOUT_RETURN_TYPE = YES_ERROR;
-				GCC_WARN_UNDECLARED_SELECTOR = YES;
-				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
-				GCC_WARN_UNUSED_FUNCTION = YES;
-				GCC_WARN_UNUSED_VARIABLE = YES;
-				MACOSX_DEPLOYMENT_TARGET = 10.8;
-				SDKROOT = macosx;
```
lucasderraugh: Great stuff 👍

GitUp/Application/AppDelegate.m
```diff
@@ -466,7 +462,13 @@ static CFDataRef _MessagePortCallBack(CFMessagePortRef local, SInt32 msgid, CFDa
   XLOG_DEBUG_CHECK(input);
   NSDictionary* output = [(__bridge AppDelegate*)info _processToolCommand:input];
   XLOG_DEBUG_CHECK(output);
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunguarded-availability"
+  // The deprecation for this method in the macOS 10.14 SDK marks the incorrect
+  // version for its introduction. However, it is useful to keep availability
+  // guards on in general. FB6233110
   return CFBridgingRetain([NSKeyedArchiver archivedDataWithRootObject:output]);
+#pragma clang diagnostic pop
```
lucasderraugh: So the secure encoding API isn't available in 10.13? Is that what you're saying here?

GitUp/Application/Document.m
```diff
@@ -85,7 +77,7 @@ static inline WindowModeID _WindowModeIDFromString(NSString* mode) {
     return kWindowModeID_Stashes;
   }
   XLOG_DEBUG_UNREACHABLE();
-  return nil;
+  return 0;
```
lucasderraugh: Feel like returning a window mode would be more appropriate here (especially if the values changed for some reason), but not really an issue in practice.

Xcode-Configurations/Base.xcconfig
```diff
@@ -2,15 +2,45 @@ ARCHS[sdk=macosx*] = $(ARCHS_STANDARD_64_BIT)
...
 ARCHS[sdk=iphoneos*] = $(ARCHS_STANDARD)
 
-MACOSX_DEPLOYMENT_TARGET = 10.8
+MACOSX_DEPLOYMENT_TARGET = 10.10
 IPHONEOS_DEPLOYMENT_TARGET = 8.0
 
 DEBUG_INFORMATION_FORMAT = dwarf-with-dsym
-COMBINE_HIDPI_IMAGES = YES
+TARGETED_DEVICE_FAMILY = 1,2,3,4
+COPY_PHASE_STRIP = NO
```
lucasderraugh: Why do we want this moved to the base config? Don't we want to strip symbols in the release configurations?

Xcode-Configurations/Release.xcconfig
```diff
@@ -1,6 +1,6 @@
 #include "Base"
 
-GCC_PREPROCESSOR_DEFINITIONS = NS_BLOCK_ASSERTIONS
-WARNING_CFLAGS = -Wall -Wno-nonnull -Wno-property-attribute-mismatch -Wno-unguarded-availability
-
+VALIDATE_PRODUCT = YES
+ENABLE_NS_ASSERTIONS = NO
 GCC_TREAT_WARNINGS_AS_ERRORS = YES
+WARNING_CFLAGS = -Wall -Wno-nonnull -Wno-property-attribute-mismatch -Wno-auto-import
```
lucasderraugh: What kind of warnings were we hitting with `-Wauto-import` that we want to avoid seeing?

lucasderraugh:
I think that about wraps up my questions. Sorry, there's just a lot of change here. Going forward if we can try and minimize changes where possible, it would be helpful in speeding up reviews. Obviously Dark Mode is an exception to this type of thing being an all or nothing proposition.

GitUp/GitUp.xcodeproj/project.pbxproj
lucasderraugh: The annoyance here though is that the Tool target will complain when I build for the Application scheme. Is this a necessary change, seems more annoying that useful, unless I'm missing something.

GitUp/Application/AppDelegate.m
zwaldowski: No, I am not saying that. `+archivedDataWithRootObject:` has been around since 10.0. When it was deprecated in the 10.14 SDK, it was marked as `API_DEPRECATED(…, macosx(10.11, 10.14), …)`. That says the method was introduced in 10.11, which it was not. Using the newer `+archivedDataWithRootObject:requiringSecureCoding:error:` would not address the warning it adds, it's hitting on availability _in the other direction_.

GitUp/Application/Document.m
zwaldowski: I had the same thought and the same preference, but I followed the convention of other locations in the project that explicitly return 0 as a sigil for an enum conversion routine.

Xcode-Configurations/Base.xcconfig
zwaldowski: You can't run stripping on a code-signed binary (like the third-party frameworks we embed) because it breaks its code-signing, so that emits a warning without a change. `COPY_PHASE_STRIP = NO` is now the default on new Xcode projects in all configurations.

GitUp/GitUp.xcodeproj/project.pbxproj
zwaldowski: > The annoyance here though is that the Tool target will complain when I build for the Application scheme.

What warning are you seeing, I don't see one?

zwaldowski: FWIW signing Command-Line Tools is now required.

zwaldowski: EDIT: Ah, the `DEVELOPMENT_TEAM` needed to be moved to the project level. I run my branches with an additional commit changing the Development Team to mine — since I'm not a member of the team and I can't build at all otherwise — and then delete it before pushing, so I did not catch it. This will be completed in one last push. :)

Xcode-Configurations/Release.xcconfig
zwaldowski: This arose from reconciling the sample project configs with the shared config. `CLANG_ENABLE_MODULES = YES` is the default in newer Xcode projects (including all of the sample projects) and required in even newer ones. When you `-Weverything` with modules enabled, you get spammed with tens of thousands of, for example: `treating #import as an import of module 'GitUpKit'`, from `-wauto-import`. Adopting module imports in the hundreds of headers could be taken on separattely, but it was definitely not my intent to do so in this PR or any time soon.

GitUp/Application/Document.m
zwaldowski: Addressed in next push

zwaldowski:
Updated.

GitUp/GitUp.xcodeproj/project.pbxproj
zwaldowski: Done

lucasderraugh:
Thanks! I’ll take a look tonight and will likely merge in. I’m on vacation so that’s why I’m somewhat sporadic in my responses, doing code reviews at night.

zwaldowski:
Oh my goodness, please enjoy your vacation first and foremost! You'd better be doing review with a drink in hand! 🍹

PR #537, opened 2019-07-01
Fix alignment of main window search field

zwaldowski:
On 10.11 and up, the search field doesn't align with the rest of the toolbar items. Fixes that.

This is a minor quality-of-life split from #532 since it's not technically related to dark mode.

### Before

Screen Shot 2019-07-01 at 2 58 45 PM

### After

Screen Shot 2019-07-01 at 2 59 24 PM


lucasderraugh:
Wonderful!

PR #539, opened 2019-07-05
Increase the contrast in commit view prompts

zwaldowski:
Trivial aesthetic change to the top of the window when rewriting, merging, splitting, etc. The prompt now follows the window background (like the footer with Cancel / Commit), instead of having a dark background like an editable text view.

### Before

before-dark
before-light

### After

after-light
after-dark


lucasderraugh:
I’ll check this out on the weekend. A little busy this week.

lucasderraugh:
Actually wasn’t as long as I thought it was for some reason. Do we still need GIColorView (I see you’re using the NSBox recommendation instead)?

zwaldowski:
As far as this PR, the `NSBox` instances added to the xibs are new separator lines. The prompts themselves just become regular `NSView` because they are grouping content but have no fill color.

Otherwise, though, you're correct: the remaining 6 uses of `GIColorView` in the project can definitely be replaced with `NSBox`.

lucasderraugh:
Looks great!

PR #548, opened 2019-08-04
Allow to use case sensitive filesystem for source code

### LINKED ISSUE #289: project does not build on a case sensitive filesystem
Either `$(PLATFORM_NAME)` changed case with some recent version of Xcode, or the build has only worked coincidentally before.

One will observe the folder is [Libraries/libssh2/MacOSX](https://github.com/git-up/Libraries/tree/04686418418ca90581233690f8edc320c5975aa1/libssh2/MacOSX) but the `$(PLATFORM_NAME)` is `macosx`. The practical implication of this is it causes `libgit2.xcodeproj` and `GitUp.xcodeproj` both to not find `libssh2.h`, since they use `$(PLATFORM_NAME)` in the header search path.

The work-around I used was to symlink `MacOSX` to `macosx` and then the build found `libssh2.h` again, but I believe the _correct_ fix is to make all those platform directories fully lowercase, since `xcodebuild -showBuildSettings | grep PLAT` shows `AVAILABLE_PLATFORMS = appletvos appletvsimulator iphoneos iphonesimulator macosx watchos watchsimulator` and thus those with case-insensitive filesystems won't notice anything, and those with case-sensitive ones will enjoy builds succeeding.

vovkasm:
Should fix #289

Why: Apple is unstable about its platform naming scheme, so use our own
build variable GITUP_PLATFORM. GITUP_PLATFORM initialized in
Base.xcconfig for easy management and set depending on sdk.


Xcode-Configurations/Base.xcconfig
```diff
@@ -46,3 +46,7 @@ GCC_WARN_UNINITIALIZED_AUTOS = YES
 GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE
 GCC_WARN_UNUSED_FUNCTION = YES
 GCC_WARN_UNUSED_VARIABLE = YES
+
+GITUP_PLATFORM[sdk=macosx*] = MacOSX
+GITUP_PLATFORM[sdk=iphonesimulator*] = iPhoneSimulator
+GITUP_PLATFORM[sdk=iphoneos*] = iPhoneOS
```
lucasderraugh: The only thing I’d like to add is a comment above the first declaration explaining why this change is necessary. Helps if someone has to look at it later and wonder what it’s for.

zwaldowski:
Can we not canonicalize `PLATFORM_NAME` with `PLATFORM_NAME:lower`? This appears to be how names are standardized in, f.ex, the new `xcframework` format.

lucasderraugh:
@zwaldowski Didn’t know that was a thing, sounds like a much solution though.

vovkasm:
@zwaldowski Can you explain more? Do you mean to change names of folders for precompiled third party libs? Or something other?

I did't try to change names of precompiled libs in `GitUpKit/Third-Party` folder because I thought it  will be more "hard to accept" change because of:
- compatibility with various Xcode versions
- hard to apply such renames on case-insensitive fs with git

Anyway I like to see any change that will allow to build application on case-sensitive fs! )

zwaldowski:
Don’t sweat it. I understand what you’re saying.

Our [tooling](https://github.com/git-up/GitUp/blob/master/GitUpKit/Third-Party/rebuild-libmbedtls.sh#L110) is as much at fault for this inconsistency as Xcode is. I would rather not continue to hide hard-codes throughout the tree.

Sorry for adding confusion. This is fine as-is so you can be unblocked. :simple_smile:

PR #549, opened 2019-08-05
Support change theme in preference panel

wqz-leo:
Added change theme pop up button in preferences panel.
Allows user to select preferred theme at runtime.
Save selected theme to UserDefaults.
Defaults to system preferences.

## Screenshots

Screen Shot 2019-08-08 at 2 52 09 pm

## License

### DIFF
--- a/GitUp/Application/AppDelegate.m
+++ b/GitUp/Application/AppDelegate.m
@@ -93,6 +93,7 @@
     kUserDefaultsKey_FirstLaunch : @(YES),
     kUserDefaultsKey_DiffWhitespaceMode : @(kGCLiveRepositoryDiffWhitespaceMode_Normal),
     kUserDefaultsKey_ShowWelcomeWindow : @(YES),
+    kUserDefaultsKey_Theme : kTheme_Dark,
   };
   [[NSUserDefaults standardUserDefaults] registerDefaults:defaults];
 }
@@ -265,6 +266,9 @@
   }
 
   [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(_willShowRecentPopUpMenu:) name:NSPopUpButtonWillPopUpNotification object:_recentPopUpButton];
+
+  NSString* theme = [[NSUserDefaults standardUserDefaults] stringForKey:kUserDefaultsKey_Theme];
+  [self _applyTheme:theme];
 }
 
 - (void)_updatePreferencePanel {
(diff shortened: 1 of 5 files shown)


lucasderraugh:
We shouldn't be defaulting to a theme one way or the other. In general we should be matching what the system is currently defined as. Now the way I would accept and the way I have seen this work that I think is a good idea is to have 3 options.
Screen Shot 2019-08-05 at 7 58 53 AM
On a separate note, was there something about dark mode that you didn't enjoy? Or do you like it but wanted it while the rest of the system was light?

wqz-leo:
Hi @lucasderraugh ,

I enjoyed the dark mode so much! However, I feel that there might be someone who would like to continue with light mode. At least we should provide an option for them to disable dark mode. IMHO theme preference should be personal but currently there's no option to disable dark mode.

I definitely can add a third option to obey system preference, but I think providing `Light` and `Dark` should be good for now. On the other hand, users probably won't change system theme nor GitUp theme very often. Please let me know if you want me to add the third option.

lucasderraugh:
@wqz-leo I like the idea, but I really need the third option to go with system preference (as the default). As it stands in this PR, everyone is going to be defaulted to dark, which is not listening to the system at all. There is also the ability coming in Catalina for the mode to dynamically change depending on time of day (and other 3rd party apps do this today).

So, I'd love the changes, but I'm convinced we need to always respect the system by default.

GitUp/Application/AppDelegate.m
```diff
@@ -277,6 +281,16 @@ - (void)_updatePreferencePanel {
   }
 }
 
+- (void)_updateThemePopUpButton {
+  NSString* theme = [[NSUserDefaults standardUserDefaults] stringForKey:kUserDefaultsKey_Theme];
+  for (NSMenuItem* item in _themePopUpButton.menu.itemArray) {
+    if ([item.title isEqualToString:theme]) {
```
lucasderraugh: I know GitUp isn't localized in other languages, but generally speaking it's bad practice to compare user facing strings to the keys backing the data (because another language likely won't have the same word). Let's make sure we're comparing either the representedObject (see `_channelPopUpButton` menu items as an example) or you could use tag number.

```
[_channelPopUpButton.menu removeAllItems];
for (NSString* string in @[ kReleaseChannel_Stable, kReleaseChannel_Continuous ]) {
  NSMenuItem* item = [[NSMenuItem alloc] initWithTitle:NSLocalizedString(string, nil) action:NULL keyEquivalent:@""];
  item.representedObject = string;
  [_channelPopUpButton.menu addItem:item];
}
```

wqz-leo: Yes I was trying to use `representedObject` initially but it didn't work very well with storyboard. Will fix it later this week and clean up the dirty code.

GitUp/Application/Base.lproj/MainMenu.xib
```diff
@@ -248,20 +248,47 @@ You must close and reopen any opened repositories in GitUp after changing this s
                                                 </binding>
                                             </connections>
                                         </matrix>
+                                        <popUpButton verticalHuggingPriority="750" id="aYx-qQ-zM6" userLabel="Theme">
+                                            <rect key="frame" x="137" y="34" width="251" height="25"/>
```
lucasderraugh: The popup button is not baseline aligned with the "Theme:" text field. Other than that, it seems to work great!

wqz-leo: All good!

PR #551, opened 2019-08-18
Animate preference pane transitions

sfsam:
This pull request implements a crossfade animation when switching between preference panes. It doesn't add any functionality, just a little polish.

### DIFF
--- a/GitUp/Application/AppDelegate.m
+++ b/GitUp/Application/AppDelegate.m
@@ -599,7 +599,11 @@ static CFDataRef _MessagePortCallBack(CFMessagePortRef local, SInt32 msgid, CFDa
 }
 
 - (IBAction)selectPreferencePane:(id)sender {
-  [_preferencesTabView selectTabViewItemWithIdentifier:_preferencesToolbar.selectedItemIdentifier];
+  [NSAnimationContext runAnimationGroup:^(NSAnimationContext * _Nonnull context) {
+    context.allowsImplicitAnimation = YES;
+    context.duration = (sender ? 0.2 : 0);
+    [_preferencesTabView selectTabViewItemWithIdentifier:_preferencesToolbar.selectedItemIdentifier];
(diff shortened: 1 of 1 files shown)


lucasderraugh:
Hey @sfsam! Sorry I've been a bit busy lately, but I'll definitely take a look at this change. If everything looks good we can probably merge it in.

lucasderraugh:
So after trying this out, I'm not sure I actually like the transition. You can see the existing text moving in the transition which seems incorrect to me. I can't think of an existing application that has that behavior and I generally like to be true to system precedence. If you have an example of an application that behaves this way, I'm all ears. Looking at Preferences in Safari, Pixelmator Pro, Finder, etc. at best have the animation of the window, but the view itself will be hidden (0 alpha) until the animation completes. I'd be ok with a Finder Preferences like behavior if we really wanted that, but those are my thoughts.

sfsam:
Thank you for the thoughtful response. If you think the animating behavior in this PR looks worse than the current non-animating behavior, you certainly shouldn't commit it. I think it is an improvement, but that is my subjective opinion. I don't think adherence to system precedence applies here since the current behavior doesn't behave like the preference panels you referenced either. The implementation of this PR simply uses Cocoa's built-in implicit animation - nothing custom by me. I think its a subjective call on whether the PR is an improvement or not. Thanks for your consideration.

lucasderraugh:
So GitUp today already animates the window size change between tabs. The only thing that doesn't animate is the view content that gets replaced. My point is that I can't find anywhere in the System that animates the content changing. Finder's behavior for example is to alpha the view 0/hide the view immediately, then animate the window size with no content, then immediately replace the content with the new view at the end of the window's size animation. Safari works this way as well. If this change was to mimic that behavior, then I would likely take it.

sfsam:
Thanks for taking a look.

PR #564, opened 2019-08-27
Fix inappropriate use of textured round button styles

### LINKED ISSUE #562: Commit view buttons appear disabled
Hi, I noticed a glitch in the UI since updating to Gitup 1.1 (1031).
The "Unstage All" and "Commit" buttons in the Commit view appear disabled. Clicking them still works though!

zwaldowski:
Closes #562. See discussion [here](https://github.com/git-up/GitUp/issues/562#issuecomment-525366827).

### Before

Screen Shot 2019-08-27 at 1 59 06 PM
Screen Shot 2019-08-27 at 1 59 13 PM

### After

Screen Shot 2019-08-27 at 1 58 54 PM
Screen Shot 2019-08-27 at 1 58 36 PM

### DIFF
--- a/GitUp/Application/Base.lproj/Document.xib
+++ b/GitUp/Application/Base.lproj/Document.xib
@@ -229,10 +229,10 @@ Adjust settings in Show menu</string>
                     <autoresizingMask key="autoresizingMask" widthSizable="YES" flexibleMaxY="YES"/>
                 </box>
                 <popUpButton id="Fov-GV-gD7">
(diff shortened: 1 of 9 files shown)


ThinkChaos:
This comment might be better off in another issue, but I've only discovered the alt-return shortcut very recently, as I expected it to be cmd+return like in the rest of the system.
Maybe that could be changed? Maybe with a transition period where a popup would suggest cmd+return when someone presses alt+return.

Issue #302, opened 2017-04-07
Ability to pick commit message from a history


plashenkov:
Hi!

It would be nice to have the ability to select a commit message from previously entered values.

ebrannin-bw:
More specifically:  Some apps have a History button by the commit-message input that lets you pick a recent commit message to put in the commit-message input.

In SourceTree, it's the clock-rewind button next to "Commit Options..."
![image](https://user-images.githubusercontent.com/86373302/230662669-201b4bc3-da4c-4e7d-8bdd-7aaa1a58983a.png)

PR #577, opened 2019-09-07
MapView: Copy commit message action

lolgear:
- [x] Copy selected commit message has been added.
- [ ] Roughly #302

### DIFF
--- a/GitUpKit/Views/Base.lproj/GIMapViewController.xib
+++ b/GitUpKit/Views/Base.lproj/GIMapViewController.xib
@@ -111,7 +111,7 @@ DQ
                         <action selector="editSelectedCommitMessage:" target="-2" id="PrK-v7-e6I"/>
                     </connections>
                 </menuItem>
-                <menuItem title="Copy Message…" alternate="YES" keyEquivalent="e" id="LnA-LI-dEb">
+                <menuItem title="Copy Message" alternate="YES" keyEquivalent="e" id="LnA-LI-dEb">
                     <modifierMask key="keyEquivalentModifierMask" option="YES"/>
                     <connections>
                         <action selector="copySelectedCommitMessage:" target="-2" id="dDP-sd-Fqm"/>
--- a/GitUpKit/Views/Base.lproj/GIMapViewController.xib
+++ b/GitUpKit/Views/Base.lproj/GIMapViewController.xib
@@ -111,6 +111,12 @@ DQ
                         <action selector="editSelectedCommitMessage:" target="-2" id="PrK-v7-e6I"/>
                     </connections>
                 </menuItem>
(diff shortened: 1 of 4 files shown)


GitUpKit/Views/GIMapViewController.m
```diff
@@ -916,6 +916,11 @@ - (IBAction)editSelectedCommitMessage:(id)sender {
   [self editCommitMessage:commit];
 }
 
+- (IBAction)copySelectedCommitMesssage:(id)sender {
```
lucasderraugh: `Messsage` -> `Message`

lolgear: Fixed :)

GitUpKit/Views/Base.lproj/GIMapViewController.xib
```diff
@@ -111,6 +111,12 @@ DQ
                         <action selector="editSelectedCommitMessage:" target="-2" id="PrK-v7-e6I"/>
                     </connections>
                 </menuItem>
+                <menuItem title="Copy Message…" alternate="YES" keyEquivalent="e" id="LnA-LI-dEb">
```
lucasderraugh: `...` is reserved for actions where there will be something the user has to follow up on. There is no followup dialog here so the text should simply be `Copy Message`.

GitUpKit/Views/GIMapViewController+Operations.m
```diff
@@ -395,6 +395,12 @@ - (void)editCommitMessage:(GCHistoryCommit*)commit {
                           }];
 }
 
+- (void)copyCommitMessage:(GCHistoryCommit*)commit {
+  NSString* message = _CleanedUpCommitMessage(commit.message);
+  [self.windowController showOverlayWithStyle:kGIOverlayStyle_Informational format:NSLocalizedString(@"Commit message copied: \"%@\"", nil), message];
+  [[NSPasteboard generalPasteboard] setString:message forType:NSPasteboardTypeString];
```
lucasderraugh: You don't have any `-[NSPasteboard declareTypes:owner:]` call. It's unlikely to be copying anything to the pasteboard in this case. From the docs: "The type must have been declared by a previous declareTypes:owner: message."

lolgear: @lucasderraugh Heh, my bad. It is ridiculous how different macOS and iOS at some points. Yes, all applications in iOS are sandboxed, so, they can't harm user.

Mmm, I've found a bug in current release app. I think that "testing" code by others is one of the best practices to find bugs.

In this PR I was in hurry, yes. It is one of the core functions that I need ( besides drag & drop, history of file, search in commit view  and search in selected range of commits ).

lucasderraugh:
Very nice, looks great!

Issue #227, opened 2016-10-05
terminal preference to support iterm


thirtytwobits:
allow selection of the terminal to invoke when using "show in terminal". For example: iterm2/3

jt3k:
!

lucasderraugh:
Should be closed with #580. You can select between Terminal and iTerm in Preferences>Advanced.

PR #580, opened 2019-09-09
External Applications: Open in iTerm

lolgear:
- [x] #227

### DIFF
--- a/GitUp/Application/Base.lproj/MainMenu.xib
+++ b/GitUp/Application/Base.lproj/MainMenu.xib
@@ -726,6 +726,19 @@ You must close and reopen any opened repositories in GitUp after changing this s
                                     <action selector="openInTerminal:" target="-1" id="YJM-LA-kZX"/>
                                 </connections>
                             </menuItem>
+                            <menuItem title="Open In External Terminal" id="pOr-se-1Oi">
+                                <modifierMask key="keyEquivalentModifierMask"/>
+                                <menu key="submenu" title="Open In External Terminal" id="dUB-iv-1C7">
+                                    <items>
+                                        <menuItem title="iTerm" id="X2d-4B-0li">
+                                            <modifierMask key="keyEquivalentModifierMask"/>
+                                            <connections>
+                                                <action selector="openIniTerm:" target="-1" id="IPe-1p-hNB"/>
+                                            </connections>
+                                        </menuItem>
+                                    </items>
+                                </menu>
+                            </menuItem>
                             <menuItem title="Open in Finder…" id="NQb-m5-0rE">
                                 <modifierMask key="keyEquivalentModifierMask"/>
                                 <connections>
--- a/GitUp/Application/Base.lproj/MainMenu.xib
+++ b/GitUp/Application/Base.lproj/MainMenu.xib
@@ -278,13 +278,13 @@ You must close and reopen any opened repositories in GitUp after changing this s
                                     </subviews>
                                 </view>
                             </tabViewItem>
-                            <tabViewItem label="{500, 262}" identifier="advanced" id="Qdl-lQ-O4q">
+                            <tabViewItem label="{500, 310}" identifier="advanced" id="Qdl-lQ-O4q">
                                 <view key="view" ambiguous="YES" id="8CG-H0-Xhk">
                                     <rect key="frame" x="0.0" y="0.0" width="500" height="410"/>
                                     <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                                     <subviews>
                                         <popUpButton verticalHuggingPriority="750" fixedFrame="YES" translatesAutoresizingMaskIntoConstraints="NO" id="7N4-S9-iWL">
-                                            <rect key="frame" x="136" y="184" width="200" height="26"/>
+                                            <rect key="frame" x="136" y="168" width="200" height="26"/>
                                             <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
                                             <popUpButtonCell key="cell" type="push" title="&lt;RELEASE CHANNEL&gt;" bezelStyle="rounded" alignment="left" lineBreakMode="truncatingTail" state="on" borderStyle="borderAndBezel" imageScaling="proportionallyDown" inset="2" autoenablesItems="NO" selectedItem="xdx-IX-FK6" id="Sdr-Ep-kPW">
                                                 <behavior key="behavior" lightByBackground="YES" lightByGray="YES"/>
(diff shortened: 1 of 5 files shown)


lucasderraugh:
I feel like we should be doing a check for valid applications the user has installed. I wonder if it should be a setting in preferences where we could check if the user even has iTerm installed.

lucasderraugh:
Could be similar to Mail in that regard:
Screen Shot 2019-09-11 at 3 02 19 PM

lolgear:
@lucasderraugh 
Could you check latest commits?
Well, I am cheating there. I have added system icons to project. It is time to find OSS icons or asks designers for appropriate icon for `Tools` tab in `Preferences` window.

lolgear:
Cheating has been removed.

lucasderraugh:
You can use Launch Services to determine which applications can accept certain types of input (document types). In this particular case it might be hard to accept any type, but I think it would be great to check that the application at least exists. We can then get the icon of the application as well.

GitUp/Application/Base.lproj/MainMenu.xib
```diff
@@ -278,6 +278,120 @@ You must close and reopen any opened repositories in GitUp after changing this s
...
+                                                <font key="font" metaFont="menu"/>
+                                                <menu key="menu" autoenablesItems="NO" id="isf-fV-z4D">
+                                                    <items>
+                                                        <menuItem title="FileMerge" id="MFB-RI-wuw"/>
+                                                        <menuItem title="Kaleidoscope" id="AO6-Py-2o5"/>
+                                                        <menuItem title="Beyond Compare" id="Cuo-4H-a9H"/>
+                                                        <menuItem title="P4Merge" id="O2n-kY-CrU"/>
+                                                        <menuItem title="DiffMerge" id="N6F-br-PGJ"/>
+                                                        <menuItem isSeparatorItem="YES" id="rLZ-QI-beC"/>
+                                                        <menuItem title="Git Tool" id="09q-g5-uZM"/>
```
lucasderraugh: Why are these being added? Where were these created before?

lolgear: @lucasderraugh 
Ah, right. I created new tab "Tools" in preferences window.

lolgear: @lucasderraugh 
Before diff-tools options (External Diff / External Merge) were in "Advanced" tab, but it is too clumsy now with additional Terminal option.
I decided to move them to new tab "Tools". You can check it and decide if it is needed.

GitUp/Application/Document.m
```diff
@@ -1965,8 +1965,33 @@ - (IBAction)openInFinder:(id)sender {
...
+    return [NSString stringWithFormat:@"tell application \"%@\" to do script \"cd \\\"%@\\\"\"", name, _repository.workingDirectoryPath];
+  }
+  if ([name isEqualToString:GIViewController_TerminalTool_iTerm]) {
+    return [NSString stringWithFormat:
+            @"""tell application \"%@\" \n"""
+            """tell current session of current window \n"""
+            """set command to \"cd \\\"%@\\\"\" \n"""
+            """write text command \n"""
+            """end tell \n"""
+            """end tell""", name, _repository.workingDirectoryPath];
```
lucasderraugh: So I noticed that this fails if I don't already have iTerm open. We should also match Terminal behavior which is to open a new window, not using the existing.

Also, I think we should leave the 2 tabs in settings instead of having the 3. I can see how it could be useful later on, but I don't think there is so much in that advanced tab yet to warrant another tab.

Lastly, we should check for existence of these applications without blindly showing iTerm if it's not even installed.

If those 3 things are addressed, this looks pretty good to go. Thanks for the work!

lolgear: @lucasderraugh 
What is the best way to hide iTerm entry in Preferences window? I suppose that Shared User defaults controller can do it, right?

lolgear:
@lucasderraugh Could you look at latest commits? Is it a right direction to implement desired behavior?
I would like also to do the following:

- [ ] Move LaunchServicesLocator to separate .h/.m
- [ ] Move constants also to this folder and rename them with new prefix GILaunchServicesLocator.
- [ ] Add bundle identifiers of other tools.
- [ ] Add keys ( standard defaults keys ) for every diff-tool if necessary.

lucasderraugh:
Let’s not make this too complicated. For now I think it’s fine to just show Terminal and iTerm as your selectable options. Don’t worry about detecting the existence of the application or worrying about other tools. I know I’m going back on what I said earlier, but let’s get the simple functionality in for this PR and then we can deal with special handling in a future PR. So if we can just do the first 2 things I said in my last comment and leave the detection for a later PR that would be great! Thanks!

lolgear:
@lucasderraugh I've already added binding to enabled property for iTerm.

Expected behavior:
App is running - open new window.
App is not running - activate app and run command in current window.

I checked script via Apple Script app - it works as expected ( script in comment ).
But I don't know why iTerm doesn't run well via NSAppleScript. When it is in non-running state, it just open window without set command in current window.

lolgear:
@lucasderraugh Check it :)

lucasderraugh:
2 things:
- The distance from the "External Terminal" to "Remotes" is too close
Screen Shot 2019-09-30 at 1 43 51 PM

- Terminal now doesn't launch to the foreground, though iTerm does. They should have consistent behavior, coming to the foreground.

lolgear:
@lucasderraugh 
Could I split windows (MainMenu.xib) into several xibs? It is too annoying to merge xib changes.

lucasderraugh:
@lolgear I think that would be nice to do at some point, but let's try to avoid these types of changes on this PR. If we're doing that, make it a separate PR.

lolgear:
@lucasderraugh Could you check latest commits?

lucasderraugh:
Looks good. The last question I have before merging is what was changed so that the xib file has `fixedFrame="YES" translatesAutoresizingMaskIntoConstraints="NO"` all over the place?

lucasderraugh:
Looks like it's resulting in:
```
MainMenu.xib: warning: Auto Layout Localization: Views without any layout constraints may clip their content or overlap other views.
```

Please revert this back to what it was previously until we require constraints in this view. Until then, standard frames work fine considering we don't have any localizations yet.

lolgear:
@lucasderraugh 
Are you sure?
Maybe it is time to split windows in different xibs and also add constraints instead of Autoresizing masks?

lolgear:
@lucasderraugh Also, there are a lot of other warnings like comparison of ints/longs, using `assign` everywhere instead of `weak` modifier.

By the way, I just downloaded master branch of this repository and open `MainMenu.xib` in Xcode 11.
It generates a lot of these warnings. Well, I consider, that malformed outdated xibs are The root of these warnings not my updates.

lucasderraugh:
Ah, you might be right. What changes did you try to make in the xib? Was it just to add the pop up button? If that’s the case I think it’s good to merge.

lucasderraugh:
On a different note, I’ve gone through the Xcode 11 warnings for non xib things, so don’t worry about those.

lolgear:
@lucasderraugh 
Yes, I added pop up button. Also I set SharedUserDefaultsController Binding in iTerm property `Enabled`.

lucasderraugh:
Cool, I’ll merge this in today. Thanks!

lolgear:
Closes #227

Issue #298, opened 2017-04-03
Can't navigate to merged branches easily via keyboard. 


macrael:
If I have a branch that has been merged into, it's difficult to navigate down both the main and the merged branch easily with the keyboard. Perhaps allowing a combination press of down and right to navigate diagonally from the main to the merged branch would help. 

Example: 

a
|
b
| \\
| c
| |
| d
| /
e

Starting with the cursor at a, there is no way to get to c using only the keyboard. But if when b was selected I could differentiate between going for e and going for c by using [down] or [down+right] respectively, then I could navigate with the keyboard completely.

lucasderraugh:
We have recently added a basic implementation using up and down with the option ⌥ key to split to the alternative branch. This works well in simple one branch cases, might evolve it if anyone has better ideas to improve it. Closing with the merge of #581.

PR #581, opened 2019-09-15
GraphView: Merged branches nodes selection via keyboard.

lolgear:
- [x] #298 
Done via `Alt` key.

### DIFF
--- a/GitUpKit/Interface/GIGraphView.m
+++ b/GitUpKit/Interface/GIGraphView.m
@@ -378,6 +378,9 @@ static const void* _associatedObjectDataKey = &_associatedObjectDataKey;
 
 - (void)_selectSideNodeAtPosition:(NSPoint)point {
   GILayer *layer = [self findLayerAtPosition:point.y];
+  if (layer == nil) {
+    return;
+  }
   NSUInteger index = [layer.nodes indexOfObjectPassingTest:^BOOL(GINode * _Nonnull obj, NSUInteger idx, BOOL * _Nonnull stop) {
     return obj != _selectedNode && !obj.dummy;
   }];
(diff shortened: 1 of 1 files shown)


lolgear:
@lucasderraugh Awaiting for review :)

lucasderraugh:
Looks good. The only thing I hit that we should fix before merging this in is that if I hold the option key and hit up when I'm already at top of tree, I hit an assertion because there is nothing higher up. We should do the same check that we do when hitting the up key normally.

lolgear:
@lucasderraugh Could you check latest commits?
I am not sure that I have added correct guard statement.
Thanks!

lolgear:
@lucasderraugh 
By the way, you don't close related issues that are mentioned in previous PRs. Is there special reason for it?

lucasderraugh:
@lolgear I'll check it in a bit. I sneak it some spare time to review these here and there.

No special reason, it's most likely because I forgot. If you want you can put in the commit "Closes #_" and it should automatically close the issue for me :).

lolgear:
@lucasderraugh 
Okey :) 
Well, I am registered at Hacktoberfest, could you add label "Hacktoberfest" to issues in my previous PRs? :)

lucasderraugh:
"To qualify for the official limited edition Hacktoberfest shirt, you must register and make four pull requests (PRs) between October 1-31 (in any time zone). PRs can be made to any public repo on GitHub, not only the ones with issues labeled Hacktoberfest."

Doesn't sound like I need to label anything other than to encourage people to find this project and that only PRs made during October would apply.

lucasderraugh:
Looks good, I was unable to hit any issues with it. I'm not sure it handles all cases as expected, but I'm more than happy to have this until somebody feels a need to change it. It's all additive anyway. Thanks again!

lolgear:
Closes #298

lolgear:
@lucasderraugh post a comment "Closes # _" in PR, right?

Issue #419, opened 2018-02-16
Commit length markers misaligned when emoji is first 'letter'.


SenHeng:
### Here the 50 and 70 word markers when an emoji is entered right at the beginning
<img width="1276" alt="emoji" src="https://user-images.githubusercontent.com/11208360/36298208-4a061dd8-133a-11e8-80a5-0b772894e28b.png">

### Here are the 50 and 70 work markers without any issues.
<img width="720" alt="nomoji" src="https://user-images.githubusercontent.com/11208360/36298209-4a2f3100-133a-11e8-81ff-c04528695a72.png">

Markers also display fine when as long as emoji are not the first letter.

Emojis entered used the official emoji keyboard (ctrl+cmd+space ).

lucasderraugh:
This has been significantly improved in #593. Not sure if you'd feel that it's fully resolved, but it's much closer to the expectation.

MrSkwiggs:
Would like to report this is still an issue on 1.1

lucasderraugh:
Yes, we haven’t created a new release. I’m going to try pushing something to the beta channel soon if you want to target that channel.

PR #593, opened 2019-10-03
Improve char width computation

pbernery:
Char width computation didn't work as expected when the first character of the NSTextView was an emoji. The `maximumAdvancement` returned the length of the emoji, placing the margins at unexpected position.

This commit improves the computation by using a fixed character (`x` has been chosen, arbitrarily). The margin will thus be placed at 50 non-emoji characters.

This is a workaround for the issue https://github.com/git-up/GitUp/issues/419.

This is not perfect though. I believe an actual fix would be to count the number of characters per line and display a warning if the the count is greater than an expected number (by default, this would be 50 for the first line, 72 for the other lines I guess).

**Before the fix**
Capture d’écran 2019-10-03 à 14 17 09

**After the fix**
Capture d’écran 2019-10-03 à 14 18 32


lucasderraugh:
Thanks @pbernery for the contribution. This looks like a wonderful improvement! Checking it out now.

pbernery:
Awesome 😊

PR #595, opened 2019-10-08
Project: Windows separation

lolgear:
As soon as windows are mature now, they requires their own xibs.
- [x] Move windows into different xibs.
- [x] Add subclass for each window.
- [x] Move outlets to window subclass.
- [x] Sync AppDelegate methods with window subclasses.
- [x] Cleanup xibs from unnecessary items.


lucasderraugh:
@lolgear I really like this initiative, but it's way too much for me to test throughly in a single PR. Could we perhaps break each xib separation into a small 1-3 file changes max per PR. One of the goals with this app is to always be in a production ready state and I don't have enough testing bandwidth to make sure that every window works correctly with 27+ file changes. If you can break your xib separations into individual change per PR, then I think we can get the ball rolling with this. Thanks!

lolgear:
@lucasderraugh 
Heh, It is nearly what you want, but I keep changes of xibs in one commit.
Let me guess appropriate PR:

1. Add custom window class.
2. Add custom window xib.
3. Remove window from MainMenu window.
4. Remove unnecessary outlets from AppDelegate.

Would it be ok?

lucasderraugh:
@lolgear Sounds perfect. Just 27 file changes is too much to test in totality, especially because xib files are hard to read with tons of changes. But yes, extract the window from MainMenu.xib and put it in it's own. Create a class if necessary to manage it and clean up ONLY the code pertaining to that window change.

Thanks so much, I'm really not trying to be a pain, but I can only keep so much in my head at one time and I don't want this

lolgear:
@lucasderraugh 
But... in this scheme MainMenu.xib and AppDelegate will be affected in each PR :)
By the way, do you know how to copy paste NSWindow to new xib? I do it by copying MainMenu.xib, not a good practice but do it job. Selecting window doesn't work in recent Xcode :(

PR #598, opened 2019-10-12
Fix Xcode 11 warnings

### LINKED ISSUE #597: Xcode 11 introduces new warnings
Fix the warnings that were introduced with Xcode 11. Mainly moving to weak references now that our deployment target is 10.10.

lucasderraugh:


lucasderraugh:
@lolgear If you get the chance, could try out this build/review and make sure I didn't miss anything.

GitUpKit/Components/GIDiffFilesViewController.h
```diff
@@ -40,6 +40,6 @@
 @property(nonatomic, readonly) NSDictionary* conflicts;
 - (void)setDeltas:(NSArray*)deltas usingConflicts:(NSDictionary*)conflicts;
 
-@property(nonatomic, assign) GCDiffDelta* selectedDelta;
-@property(nonatomic, assign) NSArray* selectedDeltas;
+@property(nonatomic, weak) GCDiffDelta* selectedDelta;
+@property(nonatomic, weak) NSArray* selectedDeltas;
```
lolgear: @lucasderraugh Are you sure that `selectedDeltas` collection should be `weak`? 
I guess it can be `copy`, no?..

lucasderraugh: Copy semantics would mean something very different, but I could use `unsafe_unretained` as well which would be the true `assign` equivalent.

lolgear: @lucasderraugh It depends on performance only. I don't think that user data in controller takes a lot of resources.
As a trade-off we have a controller that stores its data `unsafe_unretained` and doesn't control its state. copying a collection mean that you always have a copy of collection no matter what happened. It makes sense if data flow goes up and down by notifications rather than property modifiers.
Again, if you check Quick View PR, you may see a lot of issues that I caught in case of true `assign` modifiers when property suddenly start point to released object.

lucasderraugh: For this I'm not trying to really change the behavior as I want to just make a clean move to Xcode 11. We can play with different behaviors in a different PR. In this respect, I can't see how weak would be the wrong choice here, it was likely just made to be assign because of support for older platforms that didn't support weak references (though that's just a guess). To me, `weak` seems like the appropriate move for now.

lolgear: I agree in this case.

PR #599, opened 2019-10-13
Windows Separation Remastered: Welcome

lolgear:
- [x] Welcome window has been incapsulated.
- [x] Welcome window has been moved to its own xib.
- [x] Welcome window has been removed from MainMenu xib.
- [x] App Delegate loading from nibs has been added.
- [x] App Delegate windows post setup has been added.

### DIFF
--- a/GitUp/Application/AppDelegate.m
+++ b/GitUp/Application/AppDelegate.m
@@ -388,6 +388,9 @@
   // Load theme preference
   NSString* theme = [[NSUserDefaults standardUserDefaults] stringForKey:kUserDefaultsKey_Theme];
   [self _applyTheme:theme];
+  
+  // Load xibs
+  [self _loadWindowsFromBundle];
 
 #if __ENABLE_SUDDEN_TERMINATION__
   // Enable sudden termination
--- a/GitUp/Application/AppDelegate.m
+++ b/GitUp/Application/AppDelegate.m
@@ -29,6 +29,8 @@
 #import "ToolProtocol.h"
 #import "GARawTracker.h"
 
+#import "WelcomeWindow.h"
+
 #define __ENABLE_SUDDEN_TERMINATION__ 1
 
 #define kNotificationUserInfoKey_Action @"action"  // NSString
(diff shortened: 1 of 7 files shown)


lolgear:
@lucasderraugh 
It would be pleasant, if we will do it in a row, in this case I don't need to move master to actualize project and main menu xib if some PRs change them :)

lucasderraugh:
Ok, I'll try to make them a priority.

GitUp/Application/AppDelegate.m
```diff
@@ -189,46 +191,10 @@ - (void)_openDocument:(NSMenuItem*)sender {
...
-    }
-  } else {
-    NSMenuItem* item = [[NSMenuItem alloc] initWithTitle:NSLocalizedString(@"No Repositories", nil) action:NULL keyEquivalent:@""];
-    item.enabled = NO;
-    [menu addItem:item];
-  }
-}
-
 - (void)awakeFromNib {
-  _welcomeMaxHeight = _welcomeWindow.frame.size.height;
```
lucasderraugh: Good cleanup 😄

GitUp/Application/AppDelegate.m
```diff
@@ -299,6 +263,36 @@ - (void)handleDocumentCountChanged {
...
+#pragma mark - Loading Xibs
+- (id)_loadWindowFromBundleXibWithName:(NSString *)name expectedClass:(Class)class {
+  NSBundle *mainBundle = [NSBundle mainBundle];
+  NSArray *objects = @[];
+  [mainBundle loadNibNamed:name owner:self topLevelObjects:&objects];
+  NSArray *filteredObjects = [objects filteredArrayUsingPredicate:[NSPredicate predicateWithBlock:^BOOL(id  _Nullable evaluatedObject, NSDictionary<NSString *,id> * _Nullable bindings) {
+    return [evaluatedObject isKindOfClass:class];
+  }]];
+  return filteredObjects.firstObject;
+}
```
lucasderraugh: Do we expect to use this for other classes. My ideal would be that we just have a WelcomeWindow init method that does all the xib loading work for us. Could we perhaps move this logic into that class? Other than that, I think this PR looks pretty good.

lolgear: @lucasderraugh 
In previous PR I load windows and About panel by this method.
Also, look at `-loadNibNamed` invocation. It specifies owner as `self` ( AppDelegate ). Other windows don't have too much incapsulated logic ( I leave them `as-is` with some methods in `AppDelegate` ).

lolgear:
@lucasderraugh Could you merge it?

PR #600, opened 2019-10-16
Windows Separation Remastered: About

lolgear:
- [x] About panel has been incapsulated.
- [x] About panel has been moved to its own xib.
- [x] About panel has been removed from MainMenu xib.


lolgear:
@lucasderraugh Haha, I can't imagine that I will ever resolve conflicts in Github interface especially for Xib files. But... It is super-easy in this situation! Just choose this branch everywhere 😄

lolgear:
@lucasderraugh Could you check this commit?

lucasderraugh:
@lolgear I'll be checking on this when I get home today. Sorry was gone most of the weekend.

GitUp/Application/AppDelegate.m
```diff
@@ -275,6 +277,7 @@ - (id)_loadWindowFromBundleXibWithName:(NSString *)name expectedClass:(Class)cla
 }
 
 - (void)_loadWindowsFromBundle {
+  self.aboutPanel = [self _loadWindowFromBundleXibWithName:@"About" expectedClass:NSPanel.class];
```
lucasderraugh: We should be loading these lazily on request. Can we please change the approach here to encapsulate xib loading in the class instead of having the AppDelegate have this logic?

GitUp/Application/AppDelegate.m
```diff
@@ -550,17 +553,19 @@ - (IBAction)viewIssues:(id)sender {
...
   } else {
-    _versionTextField.stringValue = [NSString stringWithFormat:NSLocalizedString(@"Version %@ (%@)", nil), [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleShortVersionString"], [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleVersion"]];
+    version = [NSString stringWithFormat:NSLocalizedString(@"Version %@ (%@)", nil), [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleShortVersionString"], [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleVersion"]];
   }
 #endif
-  _copyrightTextField.stringValue = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"NSHumanReadableCopyright"];
-  [_aboutPanel makeKeyAndOrderFront:nil];
+  self.aboutPanel.versionString = version;
+  self.aboutPanel.copyrightString = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"NSHumanReadableCopyright"];
+  [self.aboutPanel makeKeyAndOrderFront:nil];
```
lucasderraugh: Let's try and move some of these smarts into the new `AboutPanel` class rather than have the AppDelegate controlling this logic.

lucasderraugh:
Looks good, I think we can make the `AboutPanel` handle more of the logic that was handled previously from the `AppDelegate`, but other than that, it looks great!

GitUp/Application/AppDelegate.m
lolgear: @lucasderraugh Do you want to load these windows on demand in `-(void)showWindow` methods?
By the way, should we also "inject" standardDefaults into them or shared instance of defaults would be fine?

GitUp/Application/AppDelegate.m
lolgear: Do you want to move all these smarts into About window?
Yes, it is possible since this logic is all around `mainBundle`.

GitUp/Application/AppDelegate.m
lucasderraugh: Either on showWindow or overriding the accessor to instantiate the window instance 1 time.

GitUp/Application/AppDelegate.m
lucasderraugh: Preferrably. I like the AppDelegate to mainly be responsible for App level issues. So what can be moved to AboutPanel seems preferable imo.

lolgear:
@lucasderraugh 
Well, I could also include Welcome window changes to fit in new NSWindow+Loading category independent style.
Should I?
It seems reasonable to do it in this PR.

lolgear:
@lucasderraugh By the way, I am thinking about new WelcomeWindowModel which will incapsulate '_allowWelcome' variable. What do you think about it?

lolgear:
@lucasderraugh Oh, no! Welcome window changes are too fat for this PR :)

GitUp/Application/AppDelegate.m
lolgear: @lucasderraugh fixed, I guess.

GitUp/Application/AboutPanel.m
```diff
@@ -0,0 +1,33 @@
...
+#import "NSWindow+Loading.h"
+@interface AboutPanel ()
+@property(nonatomic, weak) IBOutlet NSTextField* versionTextField;
+@property(nonatomic, weak) IBOutlet NSTextField* copyrightTextField;
+@end
+
+@implementation AboutPanel
++ (instancetype)new {
+  return (AboutPanel *)[self loadWindowFromBundleXibWithName:@"AboutPanel" expectedClass:NSPanel.class];
+}
```
lucasderraugh: We shouldn't be overriding new. Override init which means that either calling new or alloc init will do the correct thing.

GitUp/Application/AppDelegate.m
```diff
@@ -550,17 +553,9 @@ - (IBAction)viewIssues:(id)sender {
...
-#else
-  if (_updatePending) {
-    _versionTextField.stringValue = NSLocalizedString(@"Update Pending", nil);
-  } else {
-    _versionTextField.stringValue = [NSString stringWithFormat:NSLocalizedString(@"Version %@ (%@)", nil), [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleShortVersionString"], [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleVersion"]];
-  }
-#endif
-  _copyrightTextField.stringValue = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"NSHumanReadableCopyright"];
-  [_aboutPanel makeKeyAndOrderFront:nil];
+  self.aboutPanel = [AboutPanel new];
```
lucasderraugh: This will always create a new `AboutPanel` which I guess is ok. I'd rather just lazy load on property access but I won't hold the review back. I'm not sure the property is necessary though if we load it every time. Also, let's stick to `alloc init` paradigm for consistency.

GitUp/Application/AppDelegate.m
lucasderraugh: Should get rid of this load with load on demand.

GitUp/GitUp.xcodeproj/project.pbxproj
```diff
@@ -517,6 +540,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				CODE_SIGN_IDENTITY = "Mac Developer";
+				DEVELOPMENT_TEAM = 942QBE43P5;
```
lucasderraugh: Can't change these.

GitUp/GitUp.xcodeproj/project.pbxproj
```diff
@@ -352,7 +365,6 @@
 					};
 					E2C338AA19F8562F00063D95 = {
 						CreatedOnToolsVersion = 6.1;
-						DevelopmentTeam = 88W3E55T4B;
```
lucasderraugh: Can't change this here either.

GitUp/Application/NSWindow+Loading.m
```diff
@@ -0,0 +1,23 @@
+//
+//  NSWindow+Loading.m
+//  Application
+//
+//  Created by Dmitry Lobanov on 22.10.2019.
+//
+
+#import "NSWindow+Loading.h"
+
+#import <AppKit/AppKit.h>
```
lucasderraugh: Already importing AppKit in header.

GitUp/Application/NSWindow+Loading.h
```diff
@@ -0,0 +1,19 @@
...
+//  NSWindow+Loading.h
+//  Application
+//
+//  Created by Dmitry Lobanov on 22.10.2019.
+//
+
+#import <AppKit/AppKit.h>
+
+
+#import <Cocoa/Cocoa.h>
```
lucasderraugh: Importing Cocoa and AppKit is redundant. Just import AppKit.

lolgear:
@lucasderraugh Could you show an example with `NSWindowController`?

lucasderraugh:
@lolgear https://www.mikeash.com/pyblog/friday-qa-2013-04-05-windows-and-window-controllers.html

Basically you create an NSWindowController subclass with xib. The xib will already be configured to have the window outlet from the controller. If you're copying the window in just make sure you connect the window from the controller. Then in the window controller subclass you simply override init like:
```
- (instancetype)init {
    return [super initWithWindowNibName: @"AboutWindowController"];
}

- (void)windowDidLoad {
    // Do any setup code here.
}
```

To show the window you simply say something like this:
```
self.aboutWindowController = [[AboutWindowController alloc] init];
[self.aboutWindowController showWindow:nil];
```

lolgear:
@lucasderraugh Like this?

lucasderraugh:
@lolgear The window controller subclass looks right. We don't need the NSWindow subclass though. Just put the properties for the textfields on the window controller instead and configure them in `windowDidLoad`. Also, make sure we get rid of the `_loadWindowsFromBundle:expectedClass:` call since you'd be loading the window twice if you were doing that. The `NSWindow+Loading` file isn't needed anymore and the AboutWindow file should probably be called `AboutWindowController(.h/.m)`. Thanks for the quick responses!

GitUp/Application/AboutPanel.m
```diff
@@ -0,0 +1,42 @@
...
+@end
+
+@implementation AboutPanelWindowController
+
+- (instancetype)init {
+  return [super initWithWindowNibName:@"AboutPanel"];
+}
+
+- (void)populateWithDataWhenUpdateIsPending:(BOOL)updatePending {
+  [self loadWindow];
```
lucasderraugh: Also, don't call `loadWindow`. If you need the window to be present, calling `self.window` is enough.

lolgear:
@lucasderraugh Hmm... Are you sure that windowController with outlets of its window is better than explicit subclass of window? 
Seems reasonable for AboutPanel, but... Hmm.. `WelcomeWindow` is too heavy and perform `not` window tasks, although...

lucasderraugh:
@lolgear The only reason to subclass NSWindow these days is you need some custom drawing or behavior that can only be accomplished with subclassing. If you're simply configuring values on a window then you're best to do that all from the controller.

lucasderraugh:
@lolgear Using NSWindow subclasses isn't necessarily wrong per sey, but for xib loading NSWindowController is definitely the recommended way for 10.10+.

GitUp/Application/AppDelegate.m
```diff
@@ -550,17 +551,9 @@ - (IBAction)viewIssues:(id)sender {
...
-    _versionTextField.stringValue = NSLocalizedString(@"Update Pending", nil);
-  } else {
-    _versionTextField.stringValue = [NSString stringWithFormat:NSLocalizedString(@"Version %@ (%@)", nil), [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleShortVersionString"], [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleVersion"]];
-  }
-#endif
-  _copyrightTextField.stringValue = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"NSHumanReadableCopyright"];
-  [_aboutPanel makeKeyAndOrderFront:nil];
+  AboutPanelWindowController *aboutPanel = [[AboutPanelWindowController alloc] init];
+  [aboutPanel populateWithDataWhenUpdateIsPending:_updatePending];
+  [aboutPanel showWindow:nil];
```
lucasderraugh: Actually, I forgot that we only ever want one instance of this window, so we'll want to make sure we're loading this windowController only once. I'd personally prefer lazy loading via property. Something like:
```
- (AboutPanelWindowController *)aboutPanelWindowController {
    if (!_aboutPanelWindowController) {
        _aboutPanelWindowController = [[AboutPanelWindowController alloc] init];
    }
    return _aboutPanelWindowController;
}
```
Then simply reference `[self.aboutPanelWindowController showWindow:nil]` to have it always be a single instance.

lucasderraugh: Also, Panel and Window are redundant. I'd say call it either `AboutWindowController` or `AboutPanelController`.

lolgear:
@lucasderraugh Done :)

lucasderraugh:
Last 2 things. Remove `NSWindow+Loading` and rename AboutPanel to AboutWindowController.

lolgear:
@lucasderraugh fixed.

lolgear:
@lucasderraugh Should I also rename `AboutPanel.xib`?

lucasderraugh:
I prefer all matching names. So preferably `AboutWindowController` throughout the project in this case.

PR #604, opened 2019-10-23
Windows Separation Remastered: Welcome Window Controller

lolgear:
- [x] Welcome window controller has been encapsulated.
- [x] Welcome window controller has been moved to its own xib.


lolgear:
@lucasderraugh Hm, something strange.
Ok, I will update my repository state on next windows. A lot of commits even after I changed branch. Hm..

lolgear:
@lucasderraugh Ah, I see. I forgot about merging master into master... My bad.

GitUp/Application/AppDelegate.m
```diff
@@ -204,8 +236,6 @@ - (void)_openDocument:(NSMenuItem*)sender {
 
 - (void)awakeFromNib {
 
-  _allowWelcome = -1;
```
lolgear: What is the reason to set `_allowWelcome` to -1? Welcome window lifecycle perfectly works without it... It can be boolean, I guess.

lolgear:
@lucasderraugh Looks good?

GitUp/Application/AppDelegate.m
lucasderraugh: It's used to indicate should show (1), explicitly closed (0), and was never activated yet (-1). You can see this if you close the window and then command-tab away and then back to the app (the welcome window won't activate itself).

GitUp/Application/AppDelegate.m
```diff
@@ -44,14 +44,14 @@
 
 @interface AppDelegate () <NSUserNotificationCenterDelegate, SUUpdaterDelegate>
 @property(nonatomic, strong) AboutWindowController *aboutWindowController;
+@property(nonatomic, strong) WelcomeWindowController *welcomeWindowController;
 @property(nonatomic, strong) WelcomeWindow* welcomeWindow;
```
lucasderraugh: Used anymore?

GitUp/Application/AppDelegate.m
```diff
@@ -69,6 +69,38 @@ - (AboutWindowController *)aboutWindowController {
...
+    };
+    
+    __weak typeof(self) weakSelf = self;
+    _welcomeWindowController.model.configureItem = ^(NSMenuItem * _Nonnull item) {
+      item.target = weakSelf;
+      item.action = @selector(_openDocument:);
+    };
+    
+    _welcomeWindowController.model.openTwitter = ^{
+      [[NSWorkspace sharedWorkspace] openURL:[NSURL URLWithString:kURL_Twitter]];
```
lucasderraugh: Let the window handle what it's buttons do. We're not injecting anything useful from the `AppDelegate` here

GitUp/Application/AppDelegate.m
```diff
@@ -69,6 +69,38 @@ - (AboutWindowController *)aboutWindowController {
...
+    
+    __weak NSUserDefaults *defaults = [NSUserDefaults standardUserDefaults];
+    _welcomeWindowController.model.getUserDefaultsShouldShow = ^BOOL{
+      return [defaults boolForKey:kUserDefaultsKey_ShowWelcomeWindow];
+    };
+    
+    __weak NSDocumentController *documentController = NSDocumentController.sharedDocumentController;
+    _welcomeWindowController.model.getRecentDocuments = ^NSArray<NSURL *> * _Nonnull{
+      return documentController.recentDocumentURLs;
+    };
```
lucasderraugh: If it's not explicitly tied to the `AppDelegate` (NSUserDefaults have `standardUserDefaults`, NSDocumentController has a `sharedDocumentController`), I'm not really convinced these blocks make the code clearer. I'd prefer the window to handle this logic rather than trying to inject from the `AppDelegate` level.

GitUp/Application/AppDelegate.m
lucasderraugh: Don't think it can without changing behavior.

GitUp/Application/AppDelegate.m
```diff
@@ -261,19 +295,6 @@ - (void)_showNotificationWithTitle:(NSString*)title action:(SEL)action message:(
...
-    if (!_welcomeWindow.visible) {
-      [_welcomeWindow makeKeyAndOrderFront:nil];
-    }
-  } else {
-    if (_welcomeWindow.visible) {
-      [_welcomeWindow orderOut:nil];
-    }
-  }
-}
-
```
lucasderraugh: Seems like we should still control the window being displayed from here if we're going to keep the logic equivalent.

GitUp/Application/AppDelegate.m
lolgear: @lucasderraugh Is it desired behavior? A bit confused, though.

GitUp/Application/AppDelegate.m
```diff
@@ -425,18 +424,19 @@ - (BOOL)applicationShouldOpenUntitledFile:(NSApplication*)sender {
...
 }
 
 - (void)applicationDidBecomeActive:(NSNotification*)notification {
-  if (_allowWelcome < 0) {
-    _allowWelcome = 1;
-  }
+//  if (_allowWelcome < 0) {
+//    _allowWelcome = 1;
+//  }
+  [self.welcomeWindowController.model setShouldShow];
```
lucasderraugh: I think we should leave the previous mechanism in place for showing and dismissing the window.

GitUp/Application/AppDelegate.m
```diff
@@ -69,6 +69,38 @@ - (AboutWindowController *)aboutWindowController {
...
+      item.action = @selector(_openDocument:);
+    };
+    
+    _welcomeWindowController.model.openTwitter = ^{
+      [[NSWorkspace sharedWorkspace] openURL:[NSURL URLWithString:kURL_Twitter]];
+    };
+    
+    __weak AppDelegate *appDelegate = self;
+    _welcomeWindowController.model.viewIssues = ^{
+      [appDelegate viewIssues:nil];
```
lucasderraugh: Here as well, just let the window's button action talk to the window controller and open a URL. No need to delegate this information back to the AppDelegate.

GitUp/Application/WelcomeWindow.m
```diff
@@ -97,33 +97,125 @@ - (void)concludeDragOperation:(id<NSDraggingInfo>)sender {
...
+
+@implementation WelcomeWindowControllerModel
+#pragma mark - Show
+- (void)setShouldShow {
+  self.shouldShow = YES;
+}
+- (void)setShouldHide {
+  self.shouldShow = NO;
+}
+@end
```
lucasderraugh: I think we can just keep the old mechanism in place of explicitly telling the window (now window controller) to show or close the window.

GitUp/Application/WelcomeWindow.m
```diff
@@ -97,33 +97,125 @@ - (void)concludeDragOperation:(id<NSDraggingInfo>)sender {
...
   while (menu.numberOfItems > 1) {
     [menu removeItemAtIndex:1];
   }
 }
+
 - (void)willShowPopUpMenu {
   [self cleanupRecentEntries];
   NSMenu* menu = self.recentPopUpButton.menu;
-  NSArray* array = self.getRecentDocuments(); // [[NSDocumentController sharedDocumentController] recentDocumentURLs];
+  NSArray* array = self.model.getRecentDocuments(); // [[NSDocumentController sharedDocumentController] recentDocumentURLs];
```
lucasderraugh: Should really just call the commented code `[[NSDocumentController sharedDocumentController] recentDocumentURLs];`. Makes it much clearer.

GitUp/Application/AppDelegate.m
lolgear: @lucasderraugh I restored this behavior.

GitUp/Application/AppDelegate.m
lolgear: @lucasderraugh How to share URLs for twitter and for issues? For example, `-viewIssues:` method used both in welcome window and in main menu. Is it good to just set URLs for welcome window? ( twitterURL and forums/issuesURL )

GitUp/Application/AppDelegate.m
lolgear: @lucasderraugh ... and goes to welcome window controller 😄 
If we want to add additional logic, which is not related to welcome window lifecycle, we could expand this method in the same way by forwarding it to appropriate controller.

GitUp/Application/WelcomeWindow.m
lolgear: @lucasderraugh Fixed.

lolgear:
@lucasderraugh Should I revert back `-handleDocumentCountChanged` method and restore its logic in AppDelegate?

I think that new logic can be simplified:

```objective-c
- (void)handleDocumentCountChanged {
  BOOL showWelcomeWindow = [NSUserDefaults.standardUserDefaults boolForKey:kUserDefaultsKey_ShowWelcomeWindow];
  if (showWelcomeWindow && (self.welcomeWindowController.model.shouldShow) && !NSDocumentController.sharedDocumentController.documents.count) {
    [self.welcomeWindowController showWindow:nil];
  } else {
    [self.welcomeWindowController close];
  }
}
```

GitUp/Application/AppDelegate.m
lucasderraugh: Ah, well if it's also called from the menu item, just let the responder chain handle that. Set the action to be targeting the First Responder object in the xib and select `viewIssues:` from the list.

lolgear:
@lucasderraugh Is it ready?

GitUp/Application/WelcomeWindow.m
```diff
@@ -150,38 +245,11 @@ - (void)willShowPopUpMenu {
...
-- (void)performClose:(id)sender {
-  [self.appDelegate closeWelcomeWindow:sender];
-}
-
-#pragma mark - Window
-- (BOOL)canBecomeKeyWindow {
-  return YES;
+#pragma mark - Actions/Twitter&Issues
+- (void)openTwitter {
+  if (self.model.twitterURL) {
```
lucasderraugh: This model URL isn't needed. Welcome window is the only one using the Twitter URL.

GitUp/Application/WelcomeWindow.m
```diff
@@ -97,49 +97,144 @@ - (void)concludeDragOperation:(id<NSDraggingInfo>)sender {
...
+- (void)windowDidLoad {
+  [super windowDidLoad];
+  [self setupUIElements];
+}
+
+#pragma mark - Configuration
+- (instancetype)configuredWithModel:(WelcomeWindowControllerModel *)model {
+  self.model = model;
+  return self;
+}
```
lucasderraugh: Just do standard initialization in the init method, don't make it a pseudo init returning instancetype:
```
- (instancetype)init {
  if (self = [self initWithWindowNibName:@"Welcome"]) {
    _model = [[WelcomeWindowControllerModel alloc] init]];
  }
  return self;
 }
```
I'm also not really convinced this model object adds anything other than complexity (it's never going to be reused anywhere else).

lolgear: @lucasderraugh Fixed.

lolgear:
@lucasderraugh Is it ready?

lolgear:
@lucasderraugh 
Hm, I still don't understand some parts of Welcome Window.
For example, why not to use [menu removeAllItems] with preserving first element?

```objective-c
- (void)cleanupRecentEntries {
  NSMenu* menu = self.recentPopUpButton.menu;
  while (menu.numberOfItems > 1) {
    [menu removeItemAtIndex:1];
  }
}
```

to

```objective-c
- (void)cleanupRecentEntries {
  NSMenu* menu = self.recentPopUpButton.menu;
  NSMenuItem* item = menu.itemArray.firstObject;
  [menu removeAllItems];
  if (item) {
    [menu addItem:item];
  }
}
```

Also, I added `_openDocumentAtURL` method to `AppDelegate`. 
Now it is possible to do the following:

```objective-c
- (BOOL)performDragOperation:(id<NSDraggingInfo>)sender {
  self.receivingDrag = NO;
  
  NSURL *url = [self draggingItems:sender].firstObject;
  
  ((WelcomeWindowController *)self.window.windowController).openDocumentAtURL(url);
//  [self.appDelegate _openDocumentAtURL:url];
  
  return YES;
}
```

Is it worth to ( and appropriate to ) retrieve `openDocumentAtURL` property by using `window.windowController` chain in view?

lucasderraugh:
@lolgear Looks good to me. Let me give it a run today and I'll probably merge it in.

lolgear:
@lucasderraugh What about `-removeAllItems` and `window.windowController`? Is it ok to add these enhancements?

lucasderraugh:
@lolgear The "correct" approach for the view would be to have a completion or delegate that the window controller could assign since it would know about the view, but the view doesn't really know about the window controller. I'm fine with the cast though. removeAllItems would be fine, as long as there is no delegate callback that is expected since removeAllItems doesn't call it.

PR #610, opened 2019-11-12
Graph View: Make performance livable for a large number of branches

zwaldowski:
For a large (>400) number of branches with "Remote Branch Tips" on, looking at the graph hangs will hang for a few seconds, as well as whenever the graph reloads.

The `minY` and `maxY` passed to the graph's `drawRect:` are used to clamp which layers get drawn, ostensibly to improve performance. However, lines do not get similarly clamped between the `minX` and `maxY`, likely because branch lines can cross great distances to connect nodes, and thus wouldn't be included by a sample. At least,

This means that for 400 lines, the same vertical subset of all 400 lines get drawn repeatedly for every tile the OS asks for in the horizontal direction. That's a bunch of repeated work.

This PR adds a very small occlusion check just before actually drawing the given line using the path that it generates. The resulting performance isn't as fast as I'd hope for, but it does make using the app livable for me again while I have "Remote Branch Tips" on (i.e., 5-10s vs. 0s with dropped frames).

### DIFF
--- a/GitUpKit/Interface/GIGraphView.m
+++ b/GitUpKit/Interface/GIGraphView.m
@@ -867,13 +867,7 @@ static inline CGFloat _SquareDistanceFromPointToLine(CGFloat x0, CGFloat y0, CGF
     visible = YES;
   }
 
-  BOOL shouldDraw = visible && ^BOOL {
-    CGRect boundingBox = CGContextGetPathBoundingBox(context);
-    boundingBox.size.width = fmax(boundingBox.size.width, 1);
-    boundingBox.size.height = fmax(boundingBox.size.height, 1);
-    return [self needsToDrawRect:boundingBox];
-  }();
-
+  BOOL shouldDraw = visible && [self needsToDrawRect:CGRectInset(CGContextGetPathBoundingBox(context), -kMainLineWidth, -kMainLineWidth)];
   if (shouldDraw) {
     XLOG_DEBUG_CHECK(!line.virtual || [[(GINode*)line.nodes[0] layer] index] == 0);
     CGContextSaveGState(context);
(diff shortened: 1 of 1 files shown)


lucasderraugh:
Sounds great! Any concerns with this change or things I should test when trying it out?

GitUpKit/Interface/GIGraphView.m
```diff
@@ -1495,46 +1515,27 @@ - (void)drawRect:(NSRect)dirtyRect {
...
-      }
-    }
-  }
-
-  // Draw lines
-  if (lines.count) {
-    CGContextSetLineJoin(context, kCGLineJoinMiter);
-
+  // Draw all lines in the drawing area
+  do {
```
lucasderraugh: Any particular purpose for the `do...while`?

zwaldowski: It's intended to demonstrate that the `SetBlendMode` calls are bracket the layers/lines enumeration. That said, it is stylistic, and thus removable. In my experience the bracketing of GState is a frequent source of bugs, though. If you'd rather not have the `do { … }`, I think I'd refactor out to a `drawLines` method instead.

zwaldowski:
> Any concerns with this change or things I should test when trying it out?

Keep an eye out for clipped or missing . Overall, though, I was aiming to mostly shuffle around existing code to yield better performance instead of changing the behavior.

GitUpKit/Interface/GIGraphView.m
lucasderraugh: It's fine. Is there any difference between just having `{ ... }`? It's fine either way. I'll give this a test when I get the chance and probably merge it in.

zwaldowski: Almost certainly just me forgetting that syntax works in C. Good callout.

zwaldowski:
I've identified a small edge case (literally). I'll outset the bounds check slightly and will push that in just a moment.

Screen Shot 2019-11-13 at 10 53 34 AM

zwaldowski:
Updated.

lucasderraugh:
@zwaldowski Sorry for taking so long on this, I've been out and sick for a while and coming back to this now. Is there a repo I can test this on that shows the performance improvements, I don't have any 400+ open branch repo, but if there is a public one I'll gladly test it.

lapfelix:
Swift has 300 branches right now: https://github.com/apple/swift

PR #611, opened 2019-11-12
Windows Separation Remastered: Clone Window Controller

lolgear:
- [x] Clone window controller has been encapsulated.
- [x] Clone window controller has been moved to its own xib.

### DIFF
--- a/GitUp/Application/CloneWindowController.m
+++ b/GitUp/Application/CloneWindowController.m
@@ -34,8 +34,8 @@
 @interface CloneWindowController ()
 @property(nonatomic, weak) IBOutlet NSTextField* urlTextField;
 @property(nonatomic, weak) IBOutlet NSButton* cloneRecursiveButton;
-@property (nonatomic, assign, readonly) BOOL urlExists;
-@property (nonatomic, assign, readonly) BOOL recursive;
+@property(nonatomic, readonly) BOOL urlExists;
+@property(nonatomic, readonly) BOOL recursive;
 @end
 
 @implementation CloneWindowController
@@ -84,13 +84,11 @@
       if ([savePanel runModal] == NSFileHandlingPanelOKButton) {
         NSString *path = savePanel.URL.path;
         completion([[CloneWindowControllerResult alloc] initWithRepositoryURL:url directoryPath:path recursive:self.recursive]);
-      }
-      else {
+      } else {
         // empty directory path or cancel button pressed.
         completion([[CloneWindowControllerResult alloc] initWithRepositoryURL:url directoryPath:nil recursive:self.recursive]);
       }
-    }
-    else {
+    } else {
       // invalid repository
       completion([[CloneWindowControllerResult alloc] initWithRepositoryURL:url directoryPath:nil recursive:self.recursive]);
     }
--- a/GitUp/Application/CloneWindowController.m
+++ b/GitUp/Application/CloneWindowController.m
@@ -1,0 +1,113 @@
+//
+//  CloneWindowController.m
+//  Application
+//
+//  Created by Dmitry Lobanov on 08.10.2019.
+//
+
+#import "CloneWindowController.h"
+#import <GitUpKit/GitUpKit.h>
+
+@interface CloneWindowControllerResult ()
+- (instancetype)initWithRepositoryURL:(NSURL *)url directoryPath:(NSString *)path recursive:(BOOL)recursive;
+@end
+
+@implementation CloneWindowControllerResult
+- (instancetype)initWithRepositoryURL:(NSURL *)url directoryPath:(NSString *)path recursive:(BOOL)recursive {
+  if (url == nil || path == nil) {
+    return nil;
+  }
+  
+  if ((self = [super init])) {
+    self.repositoryURL = url;
+    self.directoryPath = path;
+    self.recursive = recursive;
+  }
+  return self;
+}
+
+- (BOOL)invalidRepository {
+  return self.repositoryURL == nil;
+}
+
+- (BOOL)emptyDirectoryPath {
+  return self.directoryPath == nil;
+}
+@end
+
+@interface CloneWindowController ()
+@property(nonatomic, weak) IBOutlet NSTextField* urlTextField;
+@property(nonatomic, weak) IBOutlet NSButton* cloneRecursiveButton;
+@property (nonatomic, assign, readonly) BOOL urlExists;
+@property (nonatomic, assign, readonly) BOOL recursive;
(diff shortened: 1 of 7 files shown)


lolgear:
@lucasderraugh Here we go! 
Next window controller :)

lolgear:
@lucasderraugh Pre-final PR :)

lolgear:
@lucasderraugh Waiting for review :)

lolgear:
@lucasderraugh Hi! Could you check this PR?

lucasderraugh:
Sorry, I’ve been sick lately. I’ll try to look at this today.

lucasderraugh:
Back from being sick/vacation week. I'll get the PR game back up and running 😊

lolgear:
@lucasderraugh I hope you are doing well now, glad to hear you have a break 😃  
Just to be sure - it wasn't burnout, right? 😟

GitUp/Application/CloneWindowController.m
```diff
@@ -0,0 +1,109 @@
...
+- (BOOL)emptyDirectoryPath {
+  return self.directoryPath == nil;
+}
+@end
+
+@interface CloneWindowController ()
+@property(nonatomic, weak) IBOutlet NSTextField* urlTextField;
+@property(nonatomic, weak) IBOutlet NSButton* cloneRecursiveButton;
+@property (nonatomic, assign, readonly) BOOL urlExists;
+@property (nonatomic, assign, readonly) BOOL recursive;
```
lucasderraugh: `assign` is redundant for `BOOL` type. Just remove it.

GitUp/Application/CloneWindowController.m
```diff
@@ -0,0 +1,109 @@
...
+
+#pragma mark - Modal
+- (void)runModalForURL:(NSString *)url completion:(nonnull void (^)(CloneWindowControllerResult * _Nonnull))completion {
+  if (!completion) {
+    return;
+  }
+  
+  self.url = url;
+  if (self.windowLoaded) {
+    [self beforeRunInModal];
```
lucasderraugh: Wouldn't this already have been run in `windowDidLoad`?

GitUp/Application/CloneWindowController.m
```diff
@@ -0,0 +1,109 @@
...
+      if ([savePanel runModal] == NSFileHandlingPanelOKButton) {
+        NSString *path = savePanel.URL.path;
+        completion([[CloneWindowControllerResult alloc] initWithRepositoryURL:url directoryPath:path recursive:self.recursive]);
+      }
+      else {
+        // empty directory path or cancel button pressed.
+        completion([[CloneWindowControllerResult alloc] initWithRepositoryURL:url directoryPath:nil recursive:self.recursive]);
+      }
+    }
+    else {
```
lucasderraugh: Let's maintain the consistent style of `} else {`

GitUp/Application/CloneWindowController.m
lolgear: It is part of "restoration cycle".

1. Window is loaded, so, all outlets are set and we can (re)set them correctly.
2. Window is not loaded, so, we skip if-condition and wait for next line where window will be loaded.

```objective-c
  if (self.windowLoaded) { // We have outlets, so, we can set UI elements
    [self beforeRunInModal];
  }
  // If not, we will skip windowLoaded if-condition and wait for `self.window` statement below. Check -windowDidLoad method for details.
  if ([NSApp runModalForWindow:self.window] && self.urlExists) {
```

We could remove `self.windowLoaded` check  and use objective-c nil-chaining feature.
```objective-c
//TIP: We don't care if window is loaded or not. This method modifies outlets. If window is not loaded, we invoke setters on nil outlets.
[self beforeRunInModal];
if ([NSApp runModalForWindow:self.window] && self.urlExists) {
```

GitUp/Application/CloneWindowController.m
lolgear: @lucasderraugh Fixed!

GitUp/Application/CloneWindowController.m
lolgear: @lucasderraugh Fixed!

GitUp/Application/CloneWindowController.m
```diff
@@ -0,0 +1,107 @@
...
+- (void)windowDidLoad {
+  [super windowDidLoad];
+  [self beforeRunInModal];
+}
+
+#pragma mark - Actions
+- (IBAction)dismissModal:(id)sender {
+  [NSApp stopModalWithCode:[(NSButton *)sender tag]];
+  [self close];
+}
```
lucasderraugh: Is this tag being explicitly used somewhere, would be nice if we could separate this out into 2 separate methods as the tag seems arbitrary.

lolgear: @lucasderraugh 

Well, they are used implicitly if I am correctly understand how `[NSApp runModalForWindow:self.window]` works. ( Return value is `[sender tag]`, right? )

GitUp/Application/CloneWindowController.m
lolgear: @lucasderraugh What do you think about it?

GitUp/Application/CloneWindowController.m
lucasderraugh: I guess I would have thought to use the predefined NSApplication `NSModalResponse` codes. I'd argue for the `NSModalResponseContinue` code to be the success case, but I'm fine with the way it's done. I'll merge this in after work today. Thanks!

lolgear: @lucasderraugh 
Could it be done by assigning ModalResponse values to buttons except assigning this value to `.tag` property?

1. Assign value to button in xib. ( .tag property? )
2. Retrieve value from `button.tag` in `-dismissModal`.
3. Access it as status result in `[NSApp runModalForWindow]`.

```objective-c

[NSApp runModalForWindow:self.window] == NSModalResponseContinue && self.urlExists
```

lucasderraugh: I think it would be more clear to just have 2 separate actions that call into stop with the Continue and Stop codes. Tags are always so obscure when reading code to understand the expected values you have to dive into the xib. Again, I'm ok with the change though. I didn't get home until late last night so I'll do it tonight. Sorry for taking forever on this.

lolgear: @lucasderraugh I will include these changes in the last PR with Preferences window. It would be nice to cleanup all rough places. :)

Issue #615, opened 2019-12-07
Zoom - increase size of everything, or at least make font size configurable


astubbs:
I'm getting old, have to squint to see the commit details on the map view.

douglashill:
Which parts of the map view you‘d like to see larger? The branch and tag labels, or the little floating view with details of the selected commit, or the lines themselves?

natestemen:
I would love to be able to increase all font size. That said, I didn't try and zoom until I was on the quickview page. So that would be the first vote for me.

astubbs:
Sorry for the slow reply - I'm referring also to the commit message, Sha, author etc on the left hand column. Not just the patch details.
And yes this:
> branch and tag labels, or the little floating view with details of the selected commit

PR #620, opened 2019-12-14
Text size setting

douglashill:

This is a resurrection of https://github.com/git-up/GitUp/pull/283 rebased onto master. The functionality is the same.

Things not addressed, and I have no immediate plans to do anything about:

- Zooming the map. That’s mostly unrelated to the font size. This seems like it could be a lot of work.
- Changing the text size in the map view’s floating selected commit info view. Since this is independent from the rest of the map layout it seems it should be fairly easy to make it respect `GIFontSize`.
- Changing the text size of the quick view’s commit message. Again, should be straightforward to make it respect and observe the font size setting.

Because of the first two points, I think this does not fix https://github.com/git-up/GitUp/issues/615 since that is specifically asking about the map view. Doesn’t really fix https://github.com/git-up/GitUp/issues/87 either but I don’t think giving the user a full font picker is the right design and sizing the text is the most useful part of picking a custom font.

Original PR text below. Everything I wrote back then is still valid so I’ve just included it verbatim expect for updating the Preferences window screenshot.

---

This branch adds a slider to let users change the font size of the diff and commit message views.

Screenshot 2019-12-14 at 18 36 29

size changes

This addresses the font size requests in the comments of https://github.com/git-up/GitUp/issues/87 — but not changing the typeface itself which is what the original comment is about. See also the alternative PR https://github.com/git-up/GitUp/pull/265 while adds very similar functionality, including changing the typeface. I’m making this pull request because I think my branch has some advantages over other the other PR.

## Design goals

Users should be able to control how large text on computers because software authors can’t know in advance what size would be most comfortable for the user to read due to factors such as visual impairments, screen pixel density and the distance between the screen and the user.

This branch aims to lay a flexible foundation for allowing any part of GitUp to be scalable, and implements scalability in the diff and commit edit views, because I think these are the two views where readability is most important. Using the same scaling in other views is open as possible future work.

Letting the user specify a scale but the application specify the typefaces and relative sizes seems like the best trade-off between allowing the user to see the app comfortably without asking them to be the designer. Also different views have use different fonts and relative sizes. Therefore, the user does not set an explicit typeface name or numerical size and instead a slider without value labels has been added to General Preferences to set the scale. This (deliberately) results in the exact size used being hard to find.

The scale factor is stored in user defaults and this value happens to match the font size of the diff view. Users with unusual requirements for especially large (or small) text can set this to any value using the `defaults` tool.

## Implementation notes

One limitation of storing the font size in the standard user defaults is that this is global state so an app using GitUpKit that wants to show two diff views at the same time would not be able to make them use different font sizes. I don’t think this use case is particularly important.

As a framework, GitUpKit should not register default values in user defaults, and it can’t guarantee that the app using GitUpKit registers a sensible default for this key. Therefore I made GitUpKit always read the font size from user defaults wrapped in a `GIFontSize()` function which ensures the value is greater than zero.

I put the `GIFontSize` function in `GIAppKit.h/m` because that’s where other GitUpKit user defaults keys are defined. I didn’t think this fits in in `GIFunctions` because it’s only available in the Mac version of GitUpKit. Putting it in `GIFunctions` wrapped in `#if __GI_HAS_APPKIT__` would work too but makes less sense to me.

I thought it would be messy to do the `#ifdef __cplusplus` like in `GIFunctions` for this one function, so I used `FOUNDATION_EXPORT`. (My preferred way to do this would be to drop the `#ifdef __cplusplus` from `GIFunctions` and define a `GIEXPORT` used for every public function.)

Naming of the user defaults key: I mirrored `GICommitMessageViewUserDefaultKey_ShowMargins` but removed the `CommitMessageView` part because this one is more general.

I decided to store the most obvious thing in user defaults: the font size used in the diff view in points. This simplicity comes at the cost of making the slider in preferences more complicated as a linear slider would not be very good: the difference between 10 and 11 points is much more important than the difference between 18 and 19 points. A value transformer is used to map between the discrete slider value and a non-linear array of font sizes that match the preset sizes in the Mac system font picker.

### DIFF
--- a/GitUp/Application/AppDelegate.m
+++ b/GitUp/Application/AppDelegate.m
@@ -93,6 +93,7 @@
     GICommitMessageViewUserDefaultKey_ShowInvisibleCharacters : @(YES),
     GICommitMessageViewUserDefaultKey_ShowMargins : @(YES),
     GICommitMessageViewUserDefaultKey_EnableSpellChecking : @(YES),
+    GIUserDefaultKey_FontSize : @(GIDefaultFontSize),
     kUserDefaultsKey_ReleaseChannel : kReleaseChannel_Stable,
     kUserDefaultsKey_CheckInterval : @(15 * 60),
     kUserDefaultsKey_FirstLaunch : @(YES),
--- a/GitUp/Application/Base.lproj/MainMenu.xib
+++ b/GitUp/Application/Base.lproj/MainMenu.xib
@@ -15,10 +15,10 @@
         <customObject id="-3" userLabel="Application" customClass="NSObject"/>
         <window title="GitUp Preferences" allowsToolTipsWhenApplicationIsInactive="NO" autorecalculatesKeyViewLoop="NO" restorable="NO" releasedWhenClosed="NO" visibleAtLaunch="NO" frameAutosaveName="preferences" animationBehavior="default" id="3qV-q4-a6k">
             <windowStyleMask key="styleMask" titled="YES" closable="YES"/>
-            <rect key="contentRect" x="940" y="240" width="500" height="410"/>
+            <rect key="contentRect" x="940" y="240" width="500" height="504"/>
             <rect key="screenRect" x="0.0" y="0.0" width="2560" height="1440"/>
             <view key="contentView" id="uEJ-lO-PMl">
-                <rect key="frame" x="0.0" y="0.0" width="500" height="410"/>
+                <rect key="frame" x="0.0" y="0.0" width="500" height="504"/>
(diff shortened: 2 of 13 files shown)


lucasderraugh:
Hey @douglashill! Thanks so much for bringing this back into the main app! I'll try to get around to reviewing this this week. I'm a bit busy at work before I'm heading out but I promise I'll get around to it.

douglashill:
No hurry. It’s a busy time of year.

lucasderraugh:
After testing it out last night and looking over the code, I think it's in a good enough place to merge in. Thank you so much for bringing this back to mainline and maybe we can make it a gift for the holiday continuous release channel 🎁.

PR #629, opened 2019-12-24
Faster multiple file staging

lapfelix:
Staging/unstaging/discarding multiple files takes a long time. Seems to be O(N) where N is multiplied by 1-2 seconds, depending on the repo.

I fixed it by only calling `-[GCRepository writeRepositoryIndex:error:]` once when doing staging/discarding operations on multiple files by adding methods for these operations that take multiple files instead of one.

I tested the changes in a big repository (with thousands of commits, ~50 branches and ~20-30 total submodules) and the time it takes to stage/unstage 19 files went from ~25.5 seconds to ~2.5 seconds. The more files are modified, the bigger the performance increase is.

### DIFF
--- a/GitUpKit/Core/GCIndex.m
+++ b/GitUpKit/Core/GCIndex.m
@@ -485,11 +485,21 @@ cleanup:
 }
 
 - (BOOL)checkoutFileToWorkingDirectory:(NSString*)path fromIndex:(GCIndex*)index error:(NSError**)error {
+  return [self checkoutFilesToWorkingDirectory:@[path]
+                                     fromIndex:index
+                                         error:error];
+}
+
+- (BOOL)checkoutFilesToWorkingDirectory:(NSArray<NSString*>*)paths fromIndex:(GCIndex*)index error:(NSError**)error {
   git_checkout_options options = GIT_CHECKOUT_OPTIONS_INIT;
   options.checkout_strategy = GIT_CHECKOUT_FORCE | GIT_CHECKOUT_DONT_UPDATE_INDEX | GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH;  // There's no reason to update the index
-  options.paths.count = 1;
-  const char* filePath = GCGitPathFromFileSystemPath(path);
-  options.paths.strings = (char**)&filePath;
+  options.paths.count = paths.count;
+  options.paths.strings = (char **)malloc((paths.count + 1) * sizeof(char*));
+  for (NSUInteger i = 0; i < paths.count; i++) {
+    const char* filePath = GCGitPathFromFileSystemPath(paths[i]);
+    options.paths.strings[i] = (char*)filePath;
+  }
+  
   CALL_LIBGIT2_FUNCTION_RETURN(NO, git_checkout_index, self.private, index.private, &options);
   return YES;
 }
--- a/GitUpKit/Core/GCIndex.m
+++ b/GitUpKit/Core/GCIndex.m
@@ -485,7 +485,7 @@ cleanup:
 }
 
 - (BOOL)checkoutFileToWorkingDirectory:(NSString*)path fromIndex:(GCIndex*)index error:(NSError**)error {
-  return [self checkoutFilesToWorkingDirectory:@[path]
+  return [self checkoutFilesToWorkingDirectory:@[ path ]
                                      fromIndex:index
                                          error:error];
 }
@@ -494,12 +494,12 @@ cleanup:
   git_checkout_options options = GIT_CHECKOUT_OPTIONS_INIT;
   options.checkout_strategy = GIT_CHECKOUT_FORCE | GIT_CHECKOUT_DONT_UPDATE_INDEX | GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH;  // There's no reason to update the index
   options.paths.count = paths.count;
-  options.paths.strings = (char **)malloc((paths.count + 1) * sizeof(char*));
+  options.paths.strings = (char**)malloc((paths.count + 1) * sizeof(char*));
   for (NSUInteger i = 0; i < paths.count; i++) {
     const char* filePath = GCGitPathFromFileSystemPath(paths[i]);
     options.paths.strings[i] = (char*)filePath;
   }
-  
+
   CALL_LIBGIT2_FUNCTION_RETURN(NO, git_checkout_index, self.private, index.private, &options);
   return YES;
 }
--- a/GitUpKit/Core/GCIndex.m
+++ b/GitUpKit/Core/GCIndex.m
@@ -494,13 +494,15 @@ cleanup:
   git_checkout_options options = GIT_CHECKOUT_OPTIONS_INIT;
   options.checkout_strategy = GIT_CHECKOUT_FORCE | GIT_CHECKOUT_DONT_UPDATE_INDEX | GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH;  // There's no reason to update the index
   options.paths.count = paths.count;
-  options.paths.strings = (char**)malloc((paths.count + 1) * sizeof(char*));
+  char** pathStrings = malloc(paths.count * sizeof(char*));
+  options.paths.strings = pathStrings;
   for (NSUInteger i = 0; i < paths.count; i++) {
     const char* filePath = GCGitPathFromFileSystemPath(paths[i]);
     options.paths.strings[i] = (char*)filePath;
   }
 
   CALL_LIBGIT2_FUNCTION_RETURN(NO, git_checkout_index, self.private, index.private, &options);
+  free(pathStrings);
   return YES;
 }
 
--- a/GitUpKit/Extensions/GCRepository+Index.m
+++ b/GitUpKit/Extensions/GCRepository+Index.m
@@ -39,30 +39,30 @@
 }
 
 - (BOOL)removeFileFromIndex:(NSString*)path error:(NSError**)error {
-  return [self removeFilesFromIndex:@[path] error:error];
+  return [self removeFilesFromIndex:@[ path ] error:error];
 }
 
-- (BOOL)removeFilesFromIndex:(NSArray<NSString *> *)paths error:(NSError**)error {
+- (BOOL)removeFilesFromIndex:(NSArray<NSString*>*)paths error:(NSError**)error {
   GCIndex* index = [self readRepositoryIndex:error];
   if (index == nil) {
     return NO;
   }
-  
-  for (NSString *path in paths) {
+
+  for (NSString* path in paths) {
     if (![self removeFile:path fromIndex:index error:error] || (error && *error != nil)) {
       [self writeRepositoryIndex:index error:error];
       return false;
     }
   }
-  
+
   return [self writeRepositoryIndex:index error:error];
 }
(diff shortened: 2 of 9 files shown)


lucasderraugh:
Absolutely wonderful! I’m out for Christmas but I will gladly review this when I’m back on the 27th. 🍻

GitUpKit/Extensions/GCRepository+Index.m
```diff
@@ -39,22 +39,61 @@ - (BOOL)resetIndexToHEAD:(NSError**)error {
...
   GCIndex* index = [self readRepositoryIndex:error];
   if (index == nil) {
     return NO;
   }
-  return [self removeFile:path fromIndex:index error:error] && [self writeRepositoryIndex:index error:error];
+
+  for (NSString* path in paths) {
+    if (![self removeFile:path fromIndex:index error:error] || (error && *error != nil)) {
+      [self writeRepositoryIndex:index error:error];
+      return false;
```
lucasderraugh: `return NO`

GitUpKit/Extensions/GCRepository+Index.m
```diff
@@ -39,22 +39,61 @@ - (BOOL)resetIndexToHEAD:(NSError**)error {
...
+    }
+
+    shouldWriteRepository = true;
+  }
+
+  if (failed && shouldWriteRepository) {
+    if (shouldWriteRepository) {
+      [self writeRepositoryIndex:index error:NULL];
+    }
+    return FALSE;
```
lucasderraugh: `return NO`

GitUpKit/Extensions/GCRepository+Index.m
```diff
@@ -39,22 +39,61 @@ - (BOOL)resetIndexToHEAD:(NSError**)error {
...
+
+  BOOL failed = NO;
+  BOOL shouldWriteRepository = NO;
+  for (NSString* path in paths) {
+    if (![self addFileInWorkingDirectory:path toIndex:index error:error] || (error && *error != nil)) {
+      failed = YES;
+      break;
+    }
+
+    shouldWriteRepository = true;
```
lucasderraugh: `= YES`

GitUpKit/Utilities/GIViewController+Utilities.h
```diff
@@ -48,11 +48,14 @@ extern NSString* const GIViewController_TerminalTool_iTerm;
 - (void)discardSubmoduleAtPath:(NSString*)path resetIndex:(BOOL)resetIndex;  // Prompts user
 
 - (void)stageAllChangesForFile:(NSString*)path;
+- (void)stageAllChangesForFiles:(NSArray<NSString*>*)paths;
 - (void)stageSelectedChangesForFile:(NSString*)path oldLines:(NSIndexSet*)oldLines newLines:(NSIndexSet*)newLines;
 - (void)unstageAllChangesForFile:(NSString*)path;
+- (void)unstageAllChangesForFiles:(NSArray<NSString*>*)path;
```
lucasderraugh: `paths`

GitUpKit/Views/GIAdvancedCommitViewController.m
```diff
@@ -187,16 +187,24 @@ - (void)diffFilesViewControllerDidChangeSelection:(GIDiffFilesViewController*)co
...
       if (delta.submodule) {
         [self stageSubmoduleAtPath:delta.canonicalPath];
+        [deltas addObject:delta];
       } else {
-        [self stageAllChangesForFile:delta.canonicalPath];
+        [nonSubmoduleDeltasPaths addObject:delta.canonicalPath];
+        [nonSubmoduleDeltas addObject:delta];
       }
-      [deltas addObject:delta];
     }
```
lucasderraugh: Seems like we can just have 2 arrays, `deltas` and `nonSubmoduleDeltasPaths`. In the end we're adding all the `deltas` together but we keep track of `nonSubmoduleDeltasPaths` to stage them all in bulk. I don't see why we need `nonSubmoduleDeltas`.

GitUpKit/Views/GIAdvancedCommitViewController.m
```diff
@@ -212,16 +220,23 @@ - (void)_stageSelectedFiles:(NSArray*)selectedDeltas {
...
       if (delta.submodule) {
         [self unstageSubmoduleAtPath:delta.canonicalPath];
+        [deltas addObject:delta];
       } else {
-        [self unstageAllChangesForFile:delta.canonicalPath];
+        [nonSubmoduleDeltasPaths addObject:delta.canonicalPath];
+        [nonSubmoduleDeltas addObject:delta];
       }
-      [deltas addObject:delta];
     }
```
lucasderraugh: Same as above.

GitUpKit/Views/GIAdvancedCommitViewController.m
lapfelix: Ah yes that totally makes sense

GitUpKit/Extensions/GCRepository+Index.m
lapfelix: ugh I started this branch 3 years ago and it shows 😂 I should've caught these when reviewing though

GitUpKit/Core/GCIndex.m
```diff
@@ -485,11 +485,21 @@ - (BOOL)resetLinesInFile:(NSString*)path index:(GCIndex*)index toCommit:(GCCommi
...
+}
+
+- (BOOL)checkoutFilesToWorkingDirectory:(NSArray<NSString*>*)paths fromIndex:(GCIndex*)index error:(NSError**)error {
   git_checkout_options options = GIT_CHECKOUT_OPTIONS_INIT;
   options.checkout_strategy = GIT_CHECKOUT_FORCE | GIT_CHECKOUT_DONT_UPDATE_INDEX | GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH;  // There's no reason to update the index
-  options.paths.count = 1;
-  const char* filePath = GCGitPathFromFileSystemPath(path);
-  options.paths.strings = (char**)&filePath;
+  options.paths.count = paths.count;
+  options.paths.strings = (char**)malloc((paths.count + 1) * sizeof(char*));
```
lucasderraugh: I believe this would actually result in a memory leak. It was previously accessing the inner pointer of the string representation, no memory was being allocated and I don't believe it would be releasing it either. I also don't think there needs to be a `+ 1` here as you're giving the count information of how many `char *`s to expect so it won't be reading past the count.

lapfelix: Oooh good catch. Not sure where that +1 comes from either. I removed it and added a `free` call after we call the libgit2 function. And I verified that it was being freed properly.

lapfelix:
I've pushed the requested change and fixed a bug in the code that calls `checkoutFilesToWorkingDirectory` (where it never sent the multiple selected file delta paths) and added comments to make it easier to understand for the next person reading it.

lucasderraugh:
The speed boost is fantastic. Thanks for this!

lucasderraugh:
@lapfelix I ended up reverting this change unfortunately. There are some major issues with discarding selected files that need resolution.

lapfelix:
Oh no! Do you have any steps to reproduce the issue?

lapfelix:
Okay I see the linked issue

lucasderraugh:
Ya, my repro steps were just selecting a few files and hitting the delete key to discard. I had cases where it would not notify about the change, but I also noticed that sometimes it would discard things I didn't select, so it had to be reverted for that reason.

GitUpKit/Views/GIAdvancedCommitViewController.m
```diff
@@ -255,18 +264,31 @@ - (void)_diffFilesViewControllerDidPressDelete:(GIDiffFilesViewController*)contr
...
                                          }
                                        }
-                                       [self.repository notifyWorkingDirectoryChanged];
-                                       if (!_workdirFilesViewController.deltas.count) {
-                                         _indexActive = YES;
-                                         [self.view.window makeFirstResponder:_indexFilesViewController.preferredFirstResponder];
+
+                                       NSError* error;
+                                       if (![self discardAllChangesForFiles:selectedFiles resetIndex:NO error:&error]) {
+                                         [self.repository notifyWorkingDirectoryChanged];
```
lucasderraugh: I believe this was the issue around why it didn't notify when discarding files. This code would only notify on failures basically. The other issue where it would discard things I didn't select was the most concerning however.

lapfelix: Ugh yeah the `notifyWorkingDirectoryChanged` and `makeFirstResponder` lines definitely shouldn't be in that `if`.

I'll review the rest of the code and see if I can reproduce that otgher issue where more things were discarded than necessary.

lapfelix: I really can't reproduce the other issue (with these lines placed at the right place). I wonder if it might be related to your other issue where the wrong files were discarded though. I'll revert my fix and see if I can make it happen like this.

lapfelix: Okaaaaayyyy I've reproduced it. It's bad.

lapfelix: Looks like the issue is in the call to the libgit2 function `git_checkout_index`. Right before it's called, the repo is fine (the right files have been deleted) but it seems like it resets too many files.

lucasderraugh:
Ya, perhaps it’s a bug on their end? We’ll have to update libgit at some point.

lapfelix:
@lucasderraugh Hmmmm I have a fix which is to remove the GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH option in `checkoutFilesToWorkingDirectory:` but I'll have to figure out why it was there in the first place

lapfelix:
@lucasderraugh I've opened a PR with the changes I'm testing: https://github.com/git-up/GitUp/pull/639

lucasderraugh:
Definitely want to know what that flag does. Thanks for looking into it though!

lapfelix:
One thing we could do to work around this apparent (?) bug is remove the flag and escape the paths we pass in so file names aren't recognized as match patterns. I've tried it and it works. I'll try it with filenames that might get recognized as match patterns before committing this change.

lucasderraugh:
I'll have to look up what this match pattern stuff means 😊. One other concern I have that I'd like to test out is what happens on failure? Previously just 1 file was skipped over, but in this case does every file fail to get merged? I'm thinking of the discard case where perhaps you have submodules and you try to discard them and a few other files but one thing fails so the entire discard fails.

lapfelix:
@lucasderraugh Submodules are still handled individually so if there's an error discarding a submodule, it will happen before we batch-process the files so we keep the same behaviour in this specific file. It will abort once it hits an error discarding a submodule.

When discarding changes in multiple files (deleting them then doing a checkout), if there's an error when deleting the file (not sure how often this happens, but it could), it will stop deleting files and will abort, present an error and stay where it was (which means the files that were already successfully deleted will stay deleted). So in that case, we won't do the checkout for these already deleted files.

If it successfully deletes the files and then fails when doing the batch-checkout, I'm not sure what will happen. I guess it will fail for all of them, yes. I haven't encountered that yet.

PR #643, opened 2020-01-29
Swapped Diff Splitter "Old Commit" and "New Commit" panel order

### LINKED ISSUE #642: Split commit panel left-hand-side old and new panels should be swapped.
Currently when you go to split a commit, the old commit panel is on top and the new commit is on the bottom. This is inverted from the standard gitup tree view which shows commits in a linear order based on the tree with the HEAD of each branch at the top.
In the split view, the "old commit" is the commit that will be below the "new commit" which will be the new HEAD (assuming splitting at the head of a branch).
Intuitively, it doesn't make sense therefore, that the "old commit" panel be in the top-left position, and would be more intuitively match the outcome by putting the "old commit" panel in the bottom-left corner of the window and the "new commit" panel in the top-left corner of the window.

This change would mean, that dragging changes from the bottom panel to the top panel, would be moving changes from the commit that will remain, into a new commit on-top of the current commit -- intuitive and clear visually.

bentemple:
This PR swaps the panel order for the "Old Commit" and the "New Commit".

The "Old Commit" when visualized in gitup will be the bottom commit, so it makes sense intuitively that by having it in the bottom-left corner of the split window, and the "New Commit" in the top-left corner of the window, when I drag changes from the old commit to the new commit, I'm dragging changes from the original commit into a new commit that will be on-top of the original commit.

Fixes #642 

Test-Plan
- Launched gitup from within the xCode project.
- Verified that when splitting a commit the old commit was in the bottom left hand corner and dragging files into the new commit panel on the top-left corner of the screen worked as expected.
- Verified that when rewriting the messages of the commits, the new commit message was on top and the old commit message was on the bottom of the opened dialog matching the order of the primary split commit view.
- Split a commit with gitup successfully.
- Verified that the new commit was above the old commit in the tree as expected and that both the new and old messages were correct.

The XIB views in XCode:

XIB View

### DIFF
--- a/GitUpKit/Views/Base.lproj/GICommitSplitterViewController.xib
+++ b/GitUpKit/Views/Base.lproj/GICommitSplitterViewController.xib
@@ -46,7 +46,7 @@
                             <rect key="frame" x="127" y="8" width="855" height="16"/>
                             <autoresizingMask key="autoresizingMask" widthSizable="YES" flexibleMinY="YES"/>
                             <textFieldCell key="cell" controlSize="small" lineBreakMode="truncatingTail" allowsUndo="NO" sendsActionOnEndEditing="YES" alignment="left" title="&lt;TITLE&gt;" id="2Ta-mo-RyA">
-                                <font key="font" metaFont="controlContent"/>
+                                <font key="font" metaFont="cellTitle"/>
                                 <color key="textColor" name="labelColor" catalog="System" colorSpace="catalog"/>
                                 <color key="backgroundColor" name="textBackgroundColor" catalog="System" colorSpace="catalog"/>
                             </textFieldCell>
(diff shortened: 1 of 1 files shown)


GitUpKit/Views/Base.lproj/GICommitSplitterViewController.xib
```diff
@@ -46,7 +46,7 @@
                             <rect key="frame" x="127" y="8" width="855" height="16"/>
                             <autoresizingMask key="autoresizingMask" widthSizable="YES" flexibleMinY="YES"/>
                             <textFieldCell key="cell" controlSize="small" lineBreakMode="truncatingTail" allowsUndo="NO" sendsActionOnEndEditing="YES" alignment="left" title="&lt;TITLE&gt;" id="2Ta-mo-RyA">
-                                <font key="font" metaFont="controlContent"/>
+                                <font key="font" metaFont="cellTitle"/>
```
bentemple: feel like this shouldn't have been changed? Probably need to revert this line?

I just dragged and swapped the panels in the xib editor in xCode

lucasderraugh:
Cool, thanks @bentemple! I'll be reviewing various PRs this weekend.

GitUpKit/Views/Base.lproj/GICommitSplitterViewController.xib
lucasderraugh: Just glancing at this diff, does this result in a different font?

lucasderraugh:
Tested it out and it looks fine. I'm a bit split (no pun intended) on the change because the commit view generally goes the other way, but I see how this appears more logical so I'll merge it in. If anyone really doesn't like it they can revert it later.

GitUpKit/Views/Base.lproj/GICommitSplitterViewController.xib
bentemple: I was thinking that too. Asked a co-worker and he said it's probably just related to me using a new version of XCode. I couldn't identify any changed fonts.

bentemple:
Thanks @lucasderraugh :) yes, definitely if anyone has an issue we can just back it out. Spoke with several people in the office and they seemed to think this made sense. 🤷‍♂ won't be offended if we revert. I have a local version I am using with this change :)

GitUpKit/Views/Base.lproj/GICommitSplitterViewController.xib
lucasderraugh: Ya, I didn't notice anything that looked wrong when looking it over.

Issue #617, opened 2019-12-09
Catalina alerts user to "malicious software" and can't directly open


lucasderraugh:
I think we'll need to send through Hardening and/or Notarization before release.
<img width="532" alt="Screen Shot 2019-12-09 at 11 55 46 AM" src="https://user-images.githubusercontent.com/714282/70467868-d7f96700-1a7a-11ea-9504-488acb83f4cc.png">

hooch:
How can I download 1.1?  I can't seem to find a link.

lucasderraugh:
@hooch There is no 1.1, the server only keeps the latest copy, which is 1.1.1 at present. You can always download the latest at gitup.co (link also in ReadMe). If you're stuck on this warning, you can right-click and hold down option key while hitting the "Open" menu item. You can also go into "Security & Privacy" in System Preferences and open it there after your first attempt.

hooch:
Thanks Lucas I dug 1.1 out of the trash.  What's involved with getting Gitup notarised properly?

lucasderraugh:
@hooch  I need to update the script @swisspol originally gave me to send the binary to apple and get a notary receipt and staple it to the binary. See this [support doc](https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution/customizing_the_notarization_workflow). I find this [article](https://blog.zeplin.io/dev-journal-automate-notarizing-macos-apps-94b0b144ba9d) much more concise though.
One main thing that needs to happen though is the app needs to adopt the Hardened Runtime requirements. I don't think we'll have any issue with this, though I haven't tried. This task could be a separate PR.

hooch:
@lucasderraugh this appears to be resolved in 1.1.2?   Thank you!

lucasderraugh:
Nah, it’s still an issue on the first install. Updates via Sparkle are unaffected.

saltcedar:
I encountered the same problem now. Would prefer not to work-around Apple security, since I'm quite a novice at this stuff. Looking forward to a fix.

lucasderraugh:
Alright, finally figured out how to get our build to pass with hardening (make sure the xcconfig doesn't affect iOS). Hopefully with the next release I'll be able to send the app through notarization and we'll be on our way!

lucasderraugh:
Hmm, looks like there could be some Sparkle hurtles here as well. I'll keep working through this over time. I'd like to cut a new build for continuous this weekend so depending on the time I get we'll see how that will shape up.

[...]

PR #646, opened 2020-02-04
Enable Hardened Runtime to prepare for Notarization

lucasderraugh:
If we want to be able to notarize the app (#617) then we'll need to have the Hardened Runtime enabled. This is step 1. The other step is for me to update the scripts with `altool` and `stapler` so we can notarize and upload all in 1 script run.


lucasderraugh:
Compiles fine locally running the travis script. Great...

PR #650, opened 2020-02-06
Fix double click on header to expand window

### LINKED ISSUE #556: Take up full screen on double click of header
On most mac apps you can double click the header and it will expand to take up the full screen (not full screen mode). Double clicking again will return the app to it's pre-expanded size

Gitup should do the same.

mattia:
In the `hitTest` of the toolbar item view we need to continue the loop
iteration when we find a `NSTextField` but it is not enabled and selectable

Fixes #556

### DIFF
--- a/GitUp/Application/ToolbarItemWrapperView.m
+++ b/GitUp/Application/ToolbarItemWrapperView.m
@@ -27,6 +27,9 @@
       if (field.enabled && field.selectable) {
         return hit;
       }
+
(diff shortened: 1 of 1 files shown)


mattia:
I took a try to fix the issue as described in the discussion. Let me know if it looks ok or I have to made changes.

Thanks for your time (and a great app!)

lucasderraugh:
Looks great @mattia! Thanks for addressing this.

Issue #190, opened 2016-08-06
Search in diff view


jankuca:
It would be quite useful to be able to search the diff of an open commit. Especially when splitting large commits, it is easy to overlook related chunks of code.

comatory:
Yes please this! I don't want to use another tool just for searching commits.

Issue #68, opened 2016-03-07
Add search for files in Commit view


lolgear:

swisspol:
Related to #34

Issue #34, opened 2016-03-02
Add filtering UI in merge conflict view to handle merges with lots of files


swisspol:
The diff view controller was not optimized to render 10,000 diffs if that's what you get :confused:

- What you want starts in the `GIConflictResolverViewController` class, which itself uses a `GIDiffContentsViewController` and `GIDiffFilesViewController`.
- The contents to display is set in `-_reloadContents`, and comes from the `GCLiveRepository` instance (the `unifiedStatus` and `indexConflicts` properties).
- These properties are live updated once enabled by `GIConflictResolverViewController` using at `self.repository.statusMode = kGCLiveRepositoryStatusMode_Unified;`.
- The property live update happens in `-[GCLiveRepository _updateStatus]` where the diffs are actually computed.

So basically one approach would be to add an extra parameter to the diff APIs so allow filtering based on name or file state or whatever you want. The actual implementation is in:

    - (GCDiff*)_diffWithType:(GCDiffType)type
                 filePattern:(NSString*)filePattern
                     options:(GCDiffOptions)options
           maxInterHunkLines:(NSUInteger)maxInterHunkLines
             maxContextLines:(NSUInteger)maxContextLines
                       error:(NSError**)error
                       block:(int (^)(git_diff** outDiff, git_diff_options* diffOptions))block {
    ...
    }

Then this parameter is is turn exposed on `GCLiveRepository` and can eventually be set by `GIConflictResolverViewController` as needed.

lolgear:
@swisspol 
It actually contains `filePattern:(NSString *)filePattern.` Is it enough to filter?
By the way, can we add `Filter`-aggregation object to simplify API?

lolgear:
@lucasderraugh Could you pin this issue?
This functionality is really necessary for next release.

PR #573, opened 2019-08-30
Quick View: Show history of file

lolgear:
- [x] Quick View: Show history of file menu entry has been added.
- [ ] Tests required.

PR #669, opened 2020-04-12
Feature. Advanced Commit View: Filter files text field.

lolgear:
- [x] Live repository file pattern variable has been added.
- [x] Advanced commit view search text field has been added.

Well, I would like to introduce another missing feature: File filtering. It is especially use in large commits.

I would like to have a feedback about this PR and feedback about roadmap of this feature.

Why it is not a final release?
I would like to add filtering functionality to each view ( view controller ) that requires it. Most of view controllers are consists of DiffFiles and DiffContents view controllers. However, it is not very simple to add this functionality into these small controllers.

So, I would like to add this functionality via inheritance from GIViewController or GIComposedViewController ( new entity ).

- [ ] Add new entity GIComposedViewController.
- [ ] Inherit all "final" controllers from this new entity.
- [ ] Add filtering functionality into base class.
- [ ] Add flags to enable/disable filtering functionality.
- [ ] Add key bindings ( like alt + F or nothing + F ).

Fixes:
- [x] #34 
- [x] #68 
- [x] #190 

Further improvements:
We could add this functionality to GIDiffFilesViewController. If we do this, we could add search text field on every view that uses GIDiffFilesViewController. However, we should add also flag to indicate if we need this `searchField` or not.

### DIFF
--- a/GitUpKit/Extensions/NSView+Embedding.h
+++ b/GitUpKit/Extensions/NSView+Embedding.h
@@ -1,0 +1,16 @@
+//
+//  NSView+Embedding.h
+//  GitUpKit (OSX)
+//
+//  Created by Dmitry Lobanov on 06.04.2020.
+//
+
+#import <AppKit/AppKit.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface NSView (Embedding)
+- (void)embedView:(NSView *)view;
+@end
+
+NS_ASSUME_NONNULL_END
--- a/GitUpKit/Extensions/NSView+Embedding.m
+++ b/GitUpKit/Extensions/NSView+Embedding.m
@@ -1,0 +1,37 @@
+//
+//  NSView+Embedding.m
+//  GitUpKit (OSX)
+//
+//  Created by Dmitry Lobanov on 06.04.2020.
+//
+
+#import "NSView+Embedding.h"
+
+@implementation NSView (Embedding)
+- (void)embedView:(NSView *)view {
+  [self.class embedView:view inView:self];
+}
+
++ (void)embedView:(NSView *)view inView:(NSView *)superview {
+  [superview addSubview:view];
+  view.translatesAutoresizingMaskIntoConstraints = NO;
+  
+  if (superview != nil && view != nil) {
+    if (@available(macOS 10.11, *)) {
+      NSArray *constraints = @[
+        [view.leftAnchor constraintEqualToAnchor:superview.leftAnchor],
+        [view.topAnchor constraintEqualToAnchor:superview.topAnchor],
+        [view.bottomAnchor constraintEqualToAnchor:superview.bottomAnchor],
+        [view.rightAnchor constraintEqualToAnchor:superview.rightAnchor]
+      ];
+      [NSLayoutConstraint activateConstraints:constraints];
+    } else {
+      // Fallback on earlier versions
+      NSArray *verticalConstraints = [NSLayoutConstraint constraintsWithVisualFormat:@"V:|[view]|" options:0 metrics:nil views:@{@"view": view}];
+      NSArray *horizontalConstraints = [NSLayoutConstraint constraintsWithVisualFormat:@"H:|[view]|" options:0 metrics:nil views:@{@"view": view}];
+      NSArray *constraints = [[NSArray arrayWithArray:verticalConstraints] arrayByAddingObjectsFromArray:horizontalConstraints];
+      [NSLayoutConstraint activateConstraints:constraints];
+    }
+  }
+}
+@end
--- a/GitUpKit/Views/Base.lproj/GIAdvancedCommitViewController.xib
+++ b/GitUpKit/Views/Base.lproj/GIAdvancedCommitViewController.xib
@@ -16,6 +16,7 @@
                 <outlet property="indexHeaderView" destination="0zR-9s-ty2" id="BLn-7k-Ti2"/>
                 <outlet property="infoTextField" destination="H0U-9x-3Cl" id="xIS-An-utS"/>
                 <outlet property="messageTextView" destination="ZGk-oR-MNP" id="bmf-B6-urc"/>
+                <outlet property="searchTextField" destination="rXa-ct-fRp" id="TLE-I1-vQt"/>
                 <outlet property="stageButton" destination="nf2-5l-LHg" id="Y5g-wM-FRo"/>
                 <outlet property="unstageButton" destination="0n6-Tx-H9g" id="RRF-VA-co7"/>
                 <outlet property="view" destination="Mge-gB-T5T" id="Dk7-8C-xIa"/>
@@ -123,8 +124,28 @@
                                                                     </subviews>
                                                                 </customView>
                                                                 <customView fixedFrame="YES" translatesAutoresizingMaskIntoConstraints="NO" id="SLl-qM-6U5">
-                                                                    <rect key="frame" x="0.0" y="0.0" width="300" height="168"/>
+                                                                    <rect key="frame" x="0.0" y="26" width="300" height="142"/>
                                                                     <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
+                                                                </customView>
+                                                                <customView id="4rD-aM-CGa">
+                                                                    <rect key="frame" x="0.0" y="0.0" width="300" height="27"/>
+                                                                    <autoresizingMask key="autoresizingMask" widthSizable="YES"/>
+                                                                    <subviews>
+                                                                        <searchField wantsLayer="YES" verticalHuggingPriority="750" textCompletion="NO" translatesAutoresizingMaskIntoConstraints="NO" id="rXa-ct-fRp">
+                                                                            <rect key="frame" x="5" y="1" width="290" height="24"/>
+                                                                            <searchFieldCell key="cell" scrollable="YES" lineBreakMode="clipping" selectable="YES" editable="YES" borderStyle="bezel" usesSingleLineMode="YES" bezelStyle="round" id="AfE-YO-lmL">
+                                                                                <font key="font" metaFont="system"/>
+                                                                                <color key="textColor" name="controlTextColor" catalog="System" colorSpace="catalog"/>
+                                                                                <color key="backgroundColor" name="textBackgroundColor" catalog="System" colorSpace="catalog"/>
+                                                                            </searchFieldCell>
+                                                                        </searchField>
+                                                                    </subviews>
+                                                                    <constraints>
+                                                                        <constraint firstItem="rXa-ct-fRp" firstAttribute="leading" secondItem="4rD-aM-CGa" secondAttribute="leading" constant="5" id="33R-mP-Dv6"/>
+                                                                        <constraint firstAttribute="bottom" secondItem="rXa-ct-fRp" secondAttribute="bottom" constant="1" id="PIM-R7-rlZ"/>
+                                                                        <constraint firstItem="rXa-ct-fRp" firstAttribute="top" secondItem="4rD-aM-CGa" secondAttribute="top" constant="2" id="nxl-no-ckt"/>
+                                                                        <constraint firstAttribute="trailing" secondItem="rXa-ct-fRp" secondAttribute="trailing" constant="5" id="z48-Ps-ceO"/>
+                                                                    </constraints>
                                                                 </customView>
                                                             </subviews>
                                                         </customView>
--- a/GitUpKit/Views/Base.lproj/GIAdvancedCommitViewController.xib
+++ b/GitUpKit/Views/Base.lproj/GIAdvancedCommitViewController.xib
@@ -104,50 +104,80 @@
                                                                 </button>
                                                             </subviews>
                                                         </customView>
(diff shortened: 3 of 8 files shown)


lolgear:
@lucasderraugh Lets discuss!

lolgear:
@lucasderraugh What do you think about it?

lucasderraugh:
I’ll check it out on Tuesday.

lucasderraugh:
Very cool! Here's some of my thoughts on it:

- I think we should try to mirror the Xcode behavior where the search is in the bottom left of the section it represents not across the entire view.
Screen Shot 2020-04-14 at 4 58 32 PM
- I think we need some protection from committing while filtering. If you accidentally have filtering on it can look like you're not committing what you think you're committing. Perhaps we should have a warning if we're hiding content that would be added to the commit.

Overall, this is pretty exciting though.

lolgear:
@lucasderraugh Could you draw an example of what you want? 
Since file pattern is applied to live repository, all controllers on advanced commit view will be affected and will get updates.
So, filtering is applying to all controllers.
Left bottom corner will be confusing for user, now?

Also, we should not forget to reset search on window disappearing, because other controllers should not know about filtering.

lucasderraugh:
Something like this, similar to how Xcode filters the content on the left side:
ScreenShot

lolgear:
@lucasderraugh 
I embedded search text field after index view, so, now I can't reach message text view on `tab`. However, if I set focus on message text view and press `shift + tab`, everything works fine and focus jumps from message to search text field and up to index and finish at working directory.

How could I set `nextKeyView` / `firstResponder` for search text field? ( Yes, I set both in setup method ).

lolgear:
@lucasderraugh 
Could you help with these two bugs:

1. I can't set any `nextKeyView/firstResponder` to make everything works fine, searchTextField has `nextKeyView` as `WorkDirectory` no matter what I set.

2. How to filter `textDidChange:(NSNotification *)notification` for concrete `NSSearchTextField`? I see that `notification.object` is not equal to `self.searchTextField` and I could check it only for following: `[[self.searchTextField subviews].firstObject subviews] == notification.object`

lolgear:
@lucasderraugh 
Haha, I just change text handling from notification center to delegate and...
Everything works fine now :D

Could you check it?

lucasderraugh:
> 2. How to filter `textDidChange:(NSNotification *)notification` for concrete `NSSearchTextField`? I see that `notification.object` is not equal to `self.searchTextField` and I could check it only for following: `[[self.searchTextField subviews].firstObject subviews] == notification.object`

Not sure about the first issue, but the second is because the object sounding out the change is probably coming from the field editor (NSTextView). You should just query the current string value on searchField when receiving that update most likely.

lolgear:
@lucasderraugh I solved it by adding delegate, it isn't an issue anymore :)

lolgear:
@lucasderraugh Could we merge it?

lolgear:
@lucasderraugh Any updates?

lucasderraugh:
Sorry @lolgear, I have been trying to finish a project whose deadline is at the end of the month. I’ll have more time in June to work on GitUp again.

lolgear:
@lucasderraugh 
I hope that you finished project before deadline and everything is smooth now :)
Do you have time to check PR and merge it if it is ok?
Also, I hope #573 could be merged too.

lucasderraugh:
Yup, I started looking at some PRs last night. I’ll go through some yours later today.

lucasderraugh:
Alrighty, I’ll go through your stuff today :)

lucasderraugh:
Alrighty, I’ll go through your stuff today :)

GitUpKit/Views/Base.lproj/GIAdvancedCommitViewController.xib
```diff
@@ -131,10 +131,30 @@
...
+                                                                <customView fixedFrame="YES" translatesAutoresizingMaskIntoConstraints="NO" id="SLl-qM-6U5">
+                                                                    <rect key="frame" x="0.0" y="26" width="300" height="142"/>
                                                                     <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                                                                 </customView>
+                                                                <customView id="0Ii-fZ-SOT">
+                                                                    <rect key="frame" x="0.0" y="0.0" width="300" height="27"/>
+                                                                    <autoresizingMask key="autoresizingMask" widthSizable="YES"/>
+                                                                    <subviews>
+                                                                        <searchField wantsLayer="YES" verticalHuggingPriority="750" textCompletion="NO" translatesAutoresizingMaskIntoConstraints="NO" id="u4y-Fx-A8g">
```
lucasderraugh, line 142: The search field should have placeholder text "Filter". We're not really searching any of the files.

lucasderraugh (changes requested):
I'm trying to figure out if it should avoid being first responder or not. I generally like to hit tab to get down to the commit text at the bottom.

My biggest thing with this though is that it takes too long to do the diff on the commit. If I understand this correctly, we're only filtering by filenames, we aren't actually searching text in these commits. If this is the case then why do we need to recompute the diffs? If we're only going to filter by filename, then we should just do that at the table view level. We've already gone through the work of diffing everything.

lolgear:
I guess it would be a little difficult to move to table-view level. The main problem is index path as id.
Yes, we search only by file name and it is unnecessary to compute diffs.
This technique with repository file pattern allows us not to do extra work and to keep consistency as "one point of truth".

lolgear:
@lucasderraugh Any updates?
We could _merge_ this PR in case of great experience, but, of course, it should be more "lightweight" and "domain-level".
However, I did exactly what is suggested in #34 .

lolgear:
@lucasderraugh 
Could we move away from xibs?
It is impossible to rebase now.

lolgear:
@lucasderraugh Done.
Could you check and merge it?

lucasderraugh:
@lolgear I think the last thing I ask on this PR is that this is an opt-in feature and that by default we don't show any filter bar. For 99% of cases it's not really useful for me and I don't want to clutter the interface if it's not useful to people. If you can add a user default checkbox in Preferences that will toggle whether this filter field is shown or not then I can merge this in.

lolgear:
@lucasderraugh Do I need additional tab for this behavior?
Is it ok if it would be done in code rather than in xib?

lucasderraugh:
> @lucasderraugh Do I need additional tab in settings for this behavior?
> Is it ok if it will be done in code rather than in xib? ( Whole layout logic ).

Let's keep the xibs and you can add it to the "General" tab for now. I'm going to move the "Appearance" based stuff into its own tab in a bit but lets keep the changes specific for this PR.

lolgear:
@lucasderraugh Done.
Yeah, for this task it was pretty simple with "data binding" through "DefaultsController". Nice OS X pattern.

lolgear:
@lucasderraugh 
Could you merge it?

lucasderraugh:
I'll resolve the conflicts here and retest. Apologies on this one.

NikKovIos:
@lolgear could you please attach a screenshot of result?

GitUpKit/Views/GIAdvancedCommitViewController.m
```diff
@@ -113,7 +158,18 @@ - (void)repositoryStatusDidUpdate {
...
 - (void)_updateCommitButton {
-  _commitButton.enabled = _indexStatus.modified || (self.repository.state == kGCRepositoryState_Merge) || self.amendButton.state;  // Creating an empty commit is OK for a merge or when amending
+  _commitButton.enabled =
+  _indexStatus.modified
+  || (self.repository.state == kGCRepositoryState_Merge)
+  || self.amendButton.state;  // Creating an empty commit is OK for a merge or when amending
+  _commitButton.enabled = _commitButton.enabled && !self.searchInProgress;
+  
+  if (self.searchInProgress) {
+    _commitButton.toolTip = NSLocalizedString(@"Search in progress", @"");
```
NikKovIos, line 168: nit: should it be localized?

Robert-M-Muench:
Can this be finished? This is the single most important feature missing when you work with a monorepo. I have about 2700 files showing up in my worktree and it's super tedious to scroll manually.

lucasderraugh:
Yes, let me work on getting this in as it's on me at this point, I just lack the time to work on GitUp much these days. The reason it hasn't landed is because I don't want to allow filtering and committing at the same time as it's important to be able to know what is being committed. I think simply blocking commit if there is an active filter is the option I'd go with. There's also a consideration of disabling the feature in Settings that I'd like to introduce.

All this is to say let me target getting it in this week and if it's not merged you can poke me again.

Robert-M-Muench:
> I don't want to allow filtering and committing at the same time as it's important to be able to know what is being committed. I think simply blocking commit if there is an active filter is the option I'd go with.

How about adding some predefined filters like staged, changed, and untracked?

Or adding a real-time updated list that always shows what a commit would include?

And this feature should be very fast. I recognized that GU takes its time to rebuild the file list. When I switch to commit view initially it takes about 10s. When switching to another application and back, the update takes about 2s every time. That's pretty disturbing.

lucasderraugh:
@Robert-M-Muench This is a text filter, we already indicate staged and unstaged via the 2 lists. The text filter is fast as we’re not querying anything new. The long loading time when transitioning is another bug that I need to address separately.

Robert-M-Muench:
@lucasderraugh Not sure I understand.

In the commit view, I see my working directory list (about 45000 files), which includes about 2700 files/uncommitted local changes (untracked, changed, etc.).

How can I filter this list to show only changed files? Or untracked files with a specific pattern?

GitUpKit/Extensions/NSView+Embedding.h
```diff
@@ -0,0 +1,16 @@
...
+//  NSView+Embedding.h
+//  GitUpKit (OSX)
+//
+//  Created by Dmitry Lobanov on 06.04.2020.
+//
+
+#import <AppKit/AppKit.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
```
kamaraj-k, line 11: i am currently learing gitup so,i like to leave comment on this one ,anyway hello there mate!,.

GitUpKit/Extensions/NSView+Embedding.m
```diff
@@ -0,0 +1,37 @@
...
+- (void)embedView:(NSView *)view {
+  [self.class embedView:view inView:self];
+}
+
++ (void)embedView:(NSView *)view inView:(NSView *)superview {
+  [superview addSubview:view];
+  view.translatesAutoresizingMaskIntoConstraints = NO;
+  
+  if (superview != nil && view != nil) {
+    if (@available(macOS 10.11, *)) {
```
NikKovIos, line 20: nit: @lolgear think about deprecate low mac versions

GitUpKit/Views/GIAdvancedCommitViewController.m
```diff
@@ -25,9 +25,12 @@
 #import "GIColorView.h"
 #import "GIInterface.h"
 #import "GIWindowController.h"
+
```
NikKovIos, line 28: Not needed spaces between imports

GitUpKit/Views/GIAdvancedCommitViewController.m
```diff
@@ -50,6 +54,40 @@ @implementation GIAdvancedCommitViewController {
   BOOL _disableFeedback;
 }
 
+#pragma mark - Search
+- (void)setupSearch {
+  self.searchTextField.delegate = self;
+  self.searchTextField.placeholderString = NSLocalizedString(@"Filter files", nil);
```
NikKovIos, line 60: nit: don't have localization

GitUpKit/Views/GIAdvancedCommitViewController.m
```diff
@@ -113,7 +158,18 @@ - (void)repositoryStatusDidUpdate {
 }
 
 - (void)_updateCommitButton {
-  _commitButton.enabled = _indexStatus.modified || (self.repository.state == kGCRepositoryState_Merge) || self.amendButton.state;  // Creating an empty commit is OK for a merge or when amending
+  _commitButton.enabled =
+  _indexStatus.modified
+  || (self.repository.state == kGCRepositoryState_Merge)
+  || self.amendButton.state;  // Creating an empty commit is OK for a merge or when amending
```
NikKovIos, line 164: don't commit anything that not involved into feature. code style is subjective.

lucasderraugh:
Oh, forgot I left this one in a state where it’s on me. Let me resume fixing this branch up and addressing the commit concerns.

GitUpKit/Views/GIAdvancedCommitViewController.m
lucasderraugh: It’s fine to go that direction anyways, perhaps even encouraged. Makes any future translation easier.

GitUpKit/Extensions/NSView+Embedding.m
lucasderraugh: @NikKovIos I’ll take a pass at cleaning this up on a new branch. The original author did enough on this one for me to push it over the finish line.

lucasderraugh:
Going to close this out for now. I've had to rework too many of these files for Tahoe and I still haven't thought of a reliable way to not make this confusing and non-invasive.

May revisit the concept in the future.

PR #688, opened 2020-06-25
Fixes Sparkle re-signing when multiple certificates are available

zwaldowski:
Follows-up #674 to prevent a failure when there are multiple Apple Development certificates in the keychain. Instead of using the generic code signing name, use the exact one that has been picked by Xcode for that target.

### DIFF
--- a/GitUp/GitUp.xcodeproj/project.pbxproj
+++ b/GitUp/GitUp.xcodeproj/project.pbxproj
@@ -491,7 +491,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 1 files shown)


lucasderraugh:
Should this work fine for everyone? I'm not really sure what `EXPANDED_CODE_SIGN_IDENTITY` does in all cases.

zwaldowski:
`EXPANDED_CODE_SIGN_IDENTITY` refers to the full name of the cert being used to sign the app itself. That variable is how Xcode communicates its choice down to the actual "Sign GitUp.app" step. `EXPANDED_CODE_SIGN_IDENTITY` is used by other auxiliary signing tools, like the Swift library embedder when that's relevant.

PR #696, opened 2020-08-07
Handle differences in user environments while executing hooks

zwaldowski:
Addresses a TODO that can cause hooks to fail if they a) reference items in the `PATH` set by fish or (new macOS system default) zsh or b) lack a shebang statement.


zwaldowski:
I tested further using different shells, everyone seems to support using `-c` but not everyone supports using `-l` along with `-c`, as well as differences in behavior using `echo`. I've updated that to be more resilient.

Further, I noticed while testing using fish that behavior was different from using `git` [on the command-line](https://github.com/git/git/blob/b6a658bd00c9c29e07f833cabfc0ef12224e277a/run-command.c#L409-L413) when a script is missing a shebang. This was actually the root cause of the issue I was running into in the first place (hooks failing mysteriously), and have updated that as well. The PR description is being updated as well. Thanks for pushing me to investigate further!

lucasderraugh:
@zwaldowski Thanks for addressing the feedback.

PR #706, opened 2020-10-20
Use user parameter when searching for keychain item

ilg:
With multiple accounts on the same server, I found that GitUp would often get the wrong credentials from the keychain, returning the username and password for a different account than the one that matched the https git URL.  Looking at KeychainAccessor.m, the `user` parameter was unused and the search was based on the value currently in the in-out `username` parameter (typically empty), so the keychain item search wasn't username-specific.  Changing `accountName` to be based on `user` made the search use the value in `user`.

Once the `user` parameter was used for searching, that meant that `accountName` was no longer `NULL`, so the `username` in-out parameter wasn't being set.  Setting `username` in the `else` block fixed this.

With these changes, I see the correct credentials being used for various accounts on the same server with https git URLs.


lucasderraugh:
Hmm, ya that definitely looks like a mistake. I'll test out the changes and merge it in. Thanks for researching this!

lucasderraugh (approved):

ilg:
Thanks, @lucasderraugh.  Please let me know if there's anything I can do to help or anything that needs to change.

PR #707, opened 2020-10-21
📄 Fill in license

Matt-Gleich:
Hello! At the moment, your license is invalid because it hasn't been filled in. Thanks for making this cool app and have a good day :)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -631,7 +631,7 @@ to attach them to the start of each source file to most effectively
 state the exclusion of warranty; and each file should have at least
 the "copyright" line and a pointer to where the full notice is found.
 
-    GitUp is a git client for the future
+    The Git interface you've been missing all your life has finally arrived.
     Copyright (C) 2020 Pierre-Olivier Latour
 
     This program is free software: you can redistribute it and/or modify
--- a/LICENSE
+++ b/LICENSE
@@ -631,7 +631,7 @@ to attach them to the start of each source file to most effectively
 state the exclusion of warranty; and each file should have at least
 the "copyright" line and a pointer to where the full notice is found.
 
-    <one line to give the program's name and a brief idea of what it does.>
+    GitUp is a git client for the future
     Copyright (C) 2020 Pierre-Olivier Latour
 
     This program is free software: you can redistribute it and/or modify
--- a/LICENSE
+++ b/LICENSE
@@ -632,7 +632,7 @@ state the exclusion of warranty; and each file should have at least
 the "copyright" line and a pointer to where the full notice is found.
 
     <one line to give the program's name and a brief idea of what it does.>
-    Copyright (C) <year>  <name of author>
+    Copyright (C) 2020 Zachary Waldowski
 
     This program is free software: you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


lucasderraugh:
Thank you, Matt!

Matt-Gleich:
@lucasderraugh are you the creator of this app? Sorry if I added the wrong name!

lucasderraugh:
Nope, name was correct. I'm just the main maintainer these days.

LICENSE
```diff
@@ -631,8 +631,8 @@ to attach them to the start of each source file to most effectively
 state the exclusion of warranty; and each file should have at least
 the "copyright" line and a pointer to where the full notice is found.
 
-    <one line to give the program's name and a brief idea of what it does.>
-    Copyright (C) <year>  <name of author>
+    GitUp is a git client for the future
```
lucasderraugh, line 634: I'd rather stick with the description in the repo.

_The Git interface you've been missing all your life has finally arrived._

lucasderraugh (approved):
I think if we stick to the original repo description, we're good to go.

Matt-Gleich:
Should be all good now!

Matt-Gleich:
@lucasderraugh any update on this?

lucasderraugh (approved):

thaJeztah:
I don't think this change is correct; I think the same rules apply for GPL as for Apache licenses when applying the license.

(I definitely know it's confusing though, and often done incorrect). The LICENSE file itself should be an unmodified copy of the license; the parts that have been replaced in this pull-request are _explaining_ how the license should be applied, which is;

- attach the following notices to the program.  It is safest to attach them to the start of each source file to most effectively state the exclusion of warranty; and each file should have at least the "copyright" line and a pointer to where the full notice is found. (see https://github.com/containerd/containerd/commit/b12c3215a0af92b9244b14c8555fc028b0d1d8c1 for an example)
- If the program does terminal interaction, make it output a short notice like this when it starts in an interactive mode:

To have the license in the repository itself, generally a `NOTICE` file is added, which may contain additional information about the copyright holder(s) and  can be referred to in the copyright notice in each file.

You can also find a discussion on this topic in https://github.com/kubernetes/kubernetes-template-project/issues/3 (related to https://github.com/kubernetes/kubernetes/commit/ad2cfc6510d032b9667c1e5ddb98312131c6beb6, which reverted the commits that filled-in the license)

Note that "copyright" can be somewhat complicated; copyright is non-transferable, and each contributor to the project will have copyright on the code they contributed, unless this repository uses a CLA (https://en.wikipedia.org/wiki/Contributor_License_Agreement) that explicitly mentions transferring copyright

>  A "contributor" is a copyright holder who authorizes use under this License of the Program or a work on which the Program is based.

For that reason, many projects use some generic term, such as "The Authors" or "The `<Project>` Authors"

lucasderraugh:
@thaJeztah Hmm interesting. I guess I'll revert it as I'm not well versed on License files and I'll yield to people that are better versed.

thaJeztah:
Yes it's super confusing; I learned from either making, or having seen the same mistakes being made elsewhere (so I don't blame anyone for assuming this is how it should be done). There's still some repositories that I work with that actually need to be fixed.

It's worth reading up a bit though, or to try and find example repositories/projects. and I should mention that I mostly work with Apache licenses; there could be some differences in how to apply the license with the license used here.

PR #710, opened 2020-11-04
Fix hint for "Diff with HEAD" feature

fo2rist:
It doesn't exit by [Spacebar], but it does by [Esc]

### DIFF
--- a/GitUp/Application/en.lproj/Help.plist
+++ b/GitUp/Application/en.lproj/Help.plist
@@ -65,7 +65,7 @@ Press [Cmd-Arrow-Up] or [Cmd-Arrow-Down] to select the previous or next commit, 
 		<key>contents</key>
 		<array>
 			<string>The Diff view shows you the diff between the current HEAD of the repository and any other commit.
-Press [i] or [Spacebar] to exit.</string>
+Press [i] or [Esc] to exit.</string>
 		</array>
 		<key>link</key>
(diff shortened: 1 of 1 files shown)


lucasderraugh:
Hmm, I wonder if it ultimately should go away with Spacebar like other diff views (when you press Spacebar on a commit) but I agree this change at least reflects reality. Thanks.

lucasderraugh (approved):

lucasderraugh:
@fo2rist Where does this text actually show up in the app. I can't seem to find where this is actually displayed 😅

fo2rist:
@lucasderraugh I think I saw it in the pop-up hint (near the top of the window), that appears the first time you use the feature. Can't reproduce b/c I closed that hint once and it doesn't pop up anymore (as expected).

fo2rist:
> Hmm, I wonder if it ultimately should go away with Spacebar like other diff views (when you press Spacebar on a commit) but I agree this change at least reflects reality. Thanks.

Good point, maybe it's better to fix the navigation instead, the commit view does exit by both Spacebar and Esc. The latter IMO is the must b/c that's what people click when they don't know what to click. So it looks to me like the Esc is the default key and Space is the shortcut for the most common action ("stage" in Working Directory, "exit" in Diff View).

If you can point me to the place where the hotkeys are registered I may be able to fix that once I have time (likely in 2021)

lucasderraugh:
It's okay. For now I'll merge this in. I'm not sure that spacebar makes much sense to leave views you don't enter with space bar anyways. Thanks for this!

Issue #693, opened 2020-07-18
Apple Universal App support (arm64)


lucasderraugh:
This issue tracks progress with arm64. So far I've done a preliminary sweep and there are dependencies we have that need to be rebuilt. This will track which dependencies have been successfully rebuild using macOS arm64.

- [x] libmbedtls
- [x] libgit2
- [x] libsqlite3
- [x] libssh2
- [x] libssl (openssl support needs to merge in before we get this and ssh working: https://github.com/openssl/openssl/pull/12369)
- [x] Sparkle

There are other things that need to be built, but these are the dependencies that are mostly out of our immediate control. I'll update this issue as time goes on.

MaddTheSane:
Sparkle should be updated to arm64. Or at least buildable.

lucasderraugh:
Done in 1.2 release.

PR #713, opened 2020-11-24
Xcode 12 bare minimum

zwaldowski:
Performs the bare minimum config changes for building on Xcode 12 without warnings.

- Updates clang-format.
- Disables ASi build so archives can complete successfully until frameworks are built for it. (See #693)

I agree to the contributor agreement, etc etc.


lucasderraugh:
Good stuff.

Issue #654, opened 2020-02-18
Glitch when dismissing a commit tooltip view in the dark mode


richardtop:
<img width="476" alt="Skärmavbild 2020-02-18 kl  21 40 41" src="https://user-images.githubusercontent.com/8013017/74771790-75e98900-5297-11ea-975e-f921d200e1de.png">

Steps to reproduce:
1. Select a commit to open a "commit tooltip view"
2. Press "spacebar" to show "commit detail view" and then press "spacebar" again to dismiss it.
3. Click somewhere else on the canvas to dismiss the tooltip
4. Observe the issue as on the screenshots

lucasderraugh:
This doesn't happen every time for you does it? I'm always in Dark mode and haven't seen this.

richardtop:
Updated the steps, so that it could be reproducible consistently.

lucasderraugh:
@richardtop Steps 2 and 3 seem impossible to me. If you click elsewhere to dismiss the tooltip, the selection goes away, so hitting spacebar doesn't do anything.

richardtop:
My bad, I've made a mistake in the sequence. Fixed now.
The "spacebar" step  should be just after opening a tooltip view.

lucasderraugh:
Hmm, still doesn’t reproduce for me. I’ll keep on the lookout for it. What OS version are you running?

richardtop:
10.15.2 (19C57)

lucasderraugh:
I'm running 10.15.3, not that I expect that to change anything. I'll keep my eye out for it though.

richardtop:
I also tried on 10.15.3 and cannot consistently reproduce the issue.

lucasderraugh:
This should now be fixed in ToT. This will land in the next release. Probably in the continuous channel next week (stable the week after if all goes well).

lucasderraugh:
This should have been addressed in our 1.2 release which is now available on the continuous channel.

PR #716, opened 2020-11-24
Fixes an issue where part of the selected node border would stay visible

zwaldowski:
Fixes a math issue where the rim of a selected node would still be visible on a retina display under certain conditions.

Image of the issue (squint hard!)

Screen Shot 2020-11-24 at 5 16 39 PM

### DIFF
--- a/GitUpKit/Interface/GIGraphView.m
+++ b/GitUpKit/Interface/GIGraphView.m
@@ -57,6 +57,7 @@
 #define kSelectedOffsetX 28
 #define kSelectedCornerRadius 5.0
 #define kSelectedTipHeight 8.0
+#define kSelectedBorderWidth 2.0
 
 #define kContextualMenuOffsetX 10
 #define kContextualMenuOffsetY -10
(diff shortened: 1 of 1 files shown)


lucasderraugh:
I believe this actually addressed #654. Somebody will be happy 😀

lucasderraugh:
Is this issue here just that the pixel boundaries aren't exact? Should it be instead flooring/rounding the `y` input?

zwaldowski:
I initially thought it was just a rounding issue, and applied an alignment method on it; but it doesn’t help. It does seem instead to be that the formula doesn’t account for the border.

lucasderraugh:
@zwaldowski Did you ever find a good way to reproduce this issue? Is there a way to always reproduce this issue if you changed an aspect of how the bubble was drawn? I was just hoping to have some repro steps to test out the changes.

zwaldowski:
Sure. It happens consistently on the tip of [this repo](https://github.com/zwaldowski/semver-release-action/issues/12) in Dark Mode when Virtual Branch Tips is ON and Stale Branch Tips is ON. It appears to happen when the rect for a selection straddles one of the system’s tile sizes. It can be exacerbated by changing the selection border width (GIGraphView.m:1369), but that alone does not make it happen on every node.

GitUpKit/Interface/GIGraphView.m
```diff
@@ -74,7 +74,7 @@
 #define CONVERT_X(x) (kSpacingX + (x)*kSpacingX)
 #define CONVERT_Y(y) (kSpacingY + (y)*kSpacingY)
 #define SQUARE(x) ((x) * (x))
-#define SELECTED_NODE_BOUNDS(x, y) NSMakeRect(x - kSpacingX / 2, y - kSelectedLabelMaxHeight / 2, kSelectedLabelMaxWidth + kSpacingX / 2 + 40, kSelectedLabelMaxHeight)
+#define SELECTED_NODE_BOUNDS(x, y) NSMakeRect(x - kSpacingX / 2, y - (kSelectedLabelMaxHeight / 2) - kSubLineWidth, kSelectedLabelMaxWidth + kSpacingX / 2 + 40, kSelectedLabelMaxHeight + (kSubLineWidth * 2))
```
lucasderraugh, line 77: I think kSubLineWidth changes the width of the lines between nodes, I don't believe that controls border width. Probably shouldn't use that, but I'm fine with making a new constant that reflects this value.

lucasderraugh (changes requested):

lucasderraugh (approved):
Thanks

PR #717, opened 2020-11-25
Fix trackpad scroll performance in files list of a merge conflict

zwaldowski:
Fixes a problem where scroll performance would suffer in the files list of the merge conflict resolver for a large commit.

Among other things, AppKit’s responsive scrolling mechanism dispatches to the main dispatch queue. Merge conflict resolution services the AppKit event loop manually to emulate modality, so that the conflict is resolved synchronously without having to redesign half of the Git methods. The AppKit event loop naturally causes the main run-loop to be serviced. Servicing the main run-loop services the main dispatch queue.

However, dispatch queues are not reentrant: if the AppKit event loop is manually serviced from inside a block on the main dispatch queue, the main dispatch queue will not be serviced until the loop completes. Thus, scrolling a table view in the merge conflict resolver would hang until AppKit gives up and falls back to old-style scrolling. This can be avoided by taking care in how the merge conflict resolver opens.

### Before

before

### After

after

### DIFF
--- a/GitUpKit/Utilities/GIAppKit.h
+++ b/GitUpKit/Utilities/GIAppKit.h
@@ -31,6 +31,8 @@ extern CGFloat const GIDefaultFontSize;
 
 FOUNDATION_EXPORT CGFloat GIFontSize(void);  // Reads GIUserDefaultKey_FontSize, falling back to GIDefaultFontSize if the user defaults value is not usable.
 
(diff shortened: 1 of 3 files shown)


lapfelix:
Yesss! Thank you for this! I started using the scroll bar to scroll through long file lists haha

lucasderraugh (approved):

Issue #723, opened 2020-12-03
Update app icon for Big Sur


mdznr:
The GitUp app icon is already a rounded rectangle. Making a version where it’s not angled will make it fit better in the dock.

<img width="175" alt="Screen Shot 2020-12-03 at 11 53 25 AM" src="https://user-images.githubusercontent.com/963948/101061137-28a48280-355e-11eb-9574-7ef93debf78e.png">

lucasderraugh:
Totally open to anyone providing assets that do this. I can probably get around to it before the next release.

mdznr:
@lucasderraugh Thanks for the quick reply!

mdznr:
Put something together quickly. What do you think about this?

<img width="767" alt="GitUp in Dock Screen Shot" src="https://user-images.githubusercontent.com/963948/101070447-6064f780-3569-11eb-97ea-d4618bf0cca7.png">

[AppIcon.iconset.zip](https://github.com/git-up/GitUp/files/5638104/AppIcon.iconset.zip)
[GitUp Big Sur App Icon.sketch.zip](https://github.com/git-up/GitUp/files/5638105/GitUp.Big.Sur.App.Icon.sketch.zip)

lucasderraugh:
Looks good to me for a Big Sur-ification. If you want to make a PR that replaces the icon assets I'd be happy to merge it in. If not, I'll get to it at some point 😄

lucasderraugh:
Closing as we integrated this into our release for 1.2. Thanks for the icon @mdznr!

Issue #703, opened 2020-09-18
Should adding be green rather than blue?


jasonhan-vassar:
Hi, it seems that adding events are colored blue instead of green, and green is instead used for merges. Given that typically (both in GitHub/GitUp), inserts are green and removals are red, would it make more sense to make adds green and removals red? This would also mean making merges blue.

Sorry if it's a silly change, but I would like to hear your guys' feedback and/or rationale. Since this is more of a personal gripe rather than a feature request, please let me know which files in the source code I should look into to change the colors myself — I am willing to change it on my own GitUp builds. Thanks!

lucasderraugh:
Can you attach a photo of what you mean by this? If you're referring to the changed files colors blue represents additions, red is removals, green is modifications, and orange is merge conflicts.

jasonhan-vassar:
Sorry if I wasn’t being clear, I was essentially asking to switch the colors blue and green. I was wondering if adding would make more sense as green as it is usually that way in other Git-related programs.

lucasderraugh:
I think GitHub uses green for additions on the diffs (which we use as well), but there isn't a color associated with adding a new file for the first time. Do you have another Git program that uses this styling for their list of file changes?

jasonhan-vassar:
One program that uses green is Fork:
<img width="1373" alt="Screen Shot 2020-09-22 at 3 14 27 PM" src="https://user-images.githubusercontent.com/34594853/93927188-873fbb80-fce6-11ea-94d1-1a6de84b3998.png">

Another is the GitHub Desktop client:
<img width="971" alt="Screen Shot 2020-09-22 at 3 23 17 PM" src="https://user-images.githubusercontent.com/34594853/93927947-92471b80-fce7-11ea-803a-e1bd13e8063b.png">

Interestingly enough, both those clients also use yellow for modifications.

Could you please add a way to customize the colors in an Appearance preference tab maybe? If not, could you let me know where to look to modify these colors myself?

lucasderraugh:
We really don't want to allow too much appearance customization but I think green for additions is probably the most logical.

lucasderraugh:
Closing as this went into the 1.2 release. Thanks @jasonhan-vassar!

PR #729, opened 2020-12-20
Swap add and modify colors, update icon for Big Sur

jasonhan-vassar:
I swapped the colors associated with added files and modified files, as discussed in #703. This is because virtually every other Git client has this color scheme and it seems to make more sense than adding being blue and modifying being green. Unfortunately, it seems the icon file sizes are larger after editing them in GIMP (~300 bytes → ~3 KB) — I'm not too sure why this is the case. If someone is more experienced with this, please feel free to make changes!

I've also updated the icons for Big Sur using the icon set provided by @mdznr in #723.

### Screenshot

Screen Shot 2020-12-20 at 12 51 19 PM

#### Normal
icon_file_a

icon_file_m

#### 2x
icon_file_a@2x

icon_file_m@2x

I agree to the contributor agreement.


lucasderraugh:
Thanks, @jasonhan-vassar! I think out of all the changes this color change one is going to be what I hear the most feedback from 😅. I think it is appropriate though and probably what most people would expect at least using GitHub. I'll review it today and get this 1.2 release out before for Christmas.

PR #714, opened 2020-11-24
Modern universal build for Apple Silicon

zwaldowski:
Provides Apple Silicon-native builds of dependent frameworks and the app itself.

Completes #693. Includes #713.

- Builds libsqlite3 3.22.0, libcrypto 1.1.1h (previously 1.1.0g; that version did not support Apple Silicon with ASM optimization), libssh 1.9.0 (previously 1.8.0; that version did not support the newer OpenSSL) for Apple Silicon.
- Updates Sparkle to a recent version supporting Apple Silicon.
- Updates build scripts to produce xcframeworks. The app’s Xcode project no longer needs to pick out a framework for itself and, in the future, will fail with more clarity on missing architecture/platform combinations.
- Removes unused build of libmbedtls.

I recommend looking at the commits here individually: 9dfec8b isn’t meaningfully reviewable.

I agree to the contributor agreement, etc etc.

PR #731, opened 2020-12-27
Fix test compilation issues

lucasderraugh:
Resolves test compilation issues introduced in schema/architecture changes. @zwaldowski do you know if there is a better approach for including libiconv.tbd and libz.tbd? I'm not sure what changed to introduce these issues but without these flags I get the following:

```
Testing failed:
	Undefined symbol: _inflateReset
	Undefined symbol: _deflateReset
	Undefined symbol: _inflateInit_
	Undefined symbol: _iconv
	Undefined symbol: _iconv_open
	Undefined symbol: _crc32
	Undefined symbol: _deflate
	Undefined symbol: _inflateEnd
	Undefined symbol: _inflate
	Undefined symbol: _deflateEnd
	Undefined symbol: _zError
	Undefined symbol: _iconv_close
	Undefined symbol: _deflateInit_
	Testing cancelled because the build failed.
```


zwaldowski:
Yeah, this totally makes sense.

Though, now that I'm looking at it, I'm confused as to why the test target manually links in a bunch of the code from the framework, rather than linking against the built framework. It just seems like asking for configuration problems.

(Anyway: I thought Travis was running on PRs, so I'm sorry #714 caused this.)

lucasderraugh:
I figured the PRs would be running these as well, but evidently that's not the case. I'll have to figure out how to set that up if I ever fix them. Without Xcode 12 I seem to fail on the tests missing some libssh2 header, but bumping it up to Xcode 12 also introduces code signing issues so I tried removing code signing but it still seems to be failing. You can see the builds here: https://travis-ci.org/github/git-up/GitUp

Issue #640, opened 2020-01-24
Previews for image assets in diff views


lucasderraugh:
GitUp doesn't show previews for image assets. We should show some kind of preview of the asset to give better context of what is being added or removed.
<img width="344" alt="Screen Shot 2020-01-24 at 12 56 31 PM" src="https://user-images.githubusercontent.com/714282/73103440-f2f82d00-3ea8-11ea-91ed-0b9d731eaede.png">

lolgear:
@lucasderraugh How to show removed images?

lucasderraugh:
If my understanding of git is correct, the image isn’t removed if it was previously staged, unless you go in and explicitly rewrite history.

PR #748, opened 2021-02-21
[Feature] Image Diff Preview

Stengo:
Partially implements https://github.com/git-up/GitUp/issues/640

This PR introduces a fairly basic image diffing cell. For now it only supports the usual 'swipe' type comparison, but more options could easily be added in the future.

### DIFF
--- a/GitUpKit/Components/GIDiffContentsViewController.m
+++ b/GitUpKit/Components/GIDiffContentsViewController.m
@@ -341,7 +341,11 @@ static NSImage* _untrackedImage = nil;
           if (patch) {
             XLOG_DEBUG_CHECK(!isBinary || patch.empty);
 
-            BOOL isImage = [[NSImage alloc] initWithContentsOfFile:[self.repository absolutePathForFile:delta.canonicalPath]] != nil;
+            CFStringRef fileExtension = CFBridgingRetain(delta.canonicalPath.pathExtension);
+            CFStringRef fileUTI = UTTypeCreatePreferredIdentifierForTag(kUTTagClassFilenameExtension, fileExtension, NULL);
+            BOOL isImage = UTTypeConformsTo(fileUTI, kUTTypeImage);
+            CFRelease(fileUTI);
+            CFRelease(fileExtension);
             if (isImage) {
               GIImageDiffView* imageDiffView = [[GIImageDiffView alloc] initWithRepository:self.repository];
               imageDiffView.delta = delta;
--- a/GitUpKit/Components/GIDiffContentsViewController.m
+++ b/GitUpKit/Components/GIDiffContentsViewController.m
@@ -56,6 +56,7 @@
 @end
 
 @interface GIImageDiffCellView : NSTableCellView
+@property(nonatomic, weak) GIImageDiffView* imageDiffView;
 @end
 
 @interface GIBinaryDiffCellView : NSTableCellView
@@ -450,10 +451,14 @@ static NSImage* _untrackedImage = nil;
     row -= 1;
   }
   if (row % 2) {
-    GITextDiffCellView* view = [rowView viewAtColumn:0];
-    if ([view isKindOfClass:[GITextDiffCellView class]]) {
-      [view.diffView removeFromSuperview];
-      view.diffView = nil;
+    GITextDiffCellView* textDiffView = [rowView viewAtColumn:0];
+    GIImageDiffCellView* imageDiffView = [rowView viewAtColumn:0];
+    if ([textDiffView isKindOfClass:[GITextDiffCellView class]]) {
+      [textDiffView.diffView removeFromSuperview];
+      textDiffView.diffView = nil;
+    } else if ([imageDiffView isKindOfClass:[GIImageDiffCellView class]]) {
+      [imageDiffView.imageDiffView removeFromSuperview];
+      imageDiffView.imageDiffView = nil;
     }
   }
 }
@@ -499,6 +504,12 @@ static inline NSString* _StringFromFileMode(GCFileMode mode) {
       return view;
     } else if (data.imageDiffView) {
       GIImageDiffCellView* view = [_tableView makeViewWithIdentifier:@"image" owner:self];
+      XLOG_DEBUG_CHECK(view.imageDiffView == nil);
+      XLOG_DEBUG_CHECK(data.imageDiffView.superview == nil);
+      data.imageDiffView.frame = view.bounds;
+      data.imageDiffView.autoresizingMask = NSViewWidthSizable | NSViewHeightSizable;
+      [view addSubview:data.imageDiffView];
+      view.imageDiffView = data.imageDiffView;
       return view;
     } else if (data.empty) {
       GIEmptyDiffCellView* view = [_tableView makeViewWithIdentifier:@"empty" owner:self];
--- a/GitUpKit/Components/GIDiffContentsViewController.m
+++ b/GitUpKit/Components/GIDiffContentsViewController.m
@@ -37,6 +37,7 @@
 @property(nonatomic, strong) GCDiffDelta* delta;
 @property(nonatomic, strong) GCIndexConflict* conflict;
 @property(nonatomic, strong) GIDiffView* diffView;
+@property(nonatomic, strong) GIImageDiffView* imageDiffView;
 @property(nonatomic, getter=isEmpty) BOOL empty;
 @end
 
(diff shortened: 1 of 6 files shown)


lucasderraugh:
I haven't tried it out but it certainly looks cool. Thanks for working on this!

gelosi:
wow! would be really cool to have it in master 🖤

lucasderraugh:
Tried the branch out and it seems pretty solid. I need to make some other changes for the 1.2.1 release before we'd but this in. This might land in 1.2.2 or 1.3 if I think it warrants the minor bump.

soberman:
Would love to see it in the app!

MrSkwiggs:
This is 🔥 . WANT

GitUpKit/Components/GIDiffContentsViewController.m
```diff
@@ -332,7 +340,13 @@ - (void)_reloadDeltas {
           GCDiffPatch* patch = [self.repository makePatchForDiffDelta:delta isBinary:&isBinary error:&error];
           if (patch) {
             XLOG_DEBUG_CHECK(!isBinary || patch.empty);
-            if (patch.empty) {
+
+            BOOL isImage = [[NSImage alloc] initWithContentsOfFile:[self.repository absolutePathForFile:delta.canonicalPath]] != nil;
```
lucasderraugh, line 344: I think we need a lighter weight approach for determining if a path may be an image. If we have a lot of images and large images then we're going to load them all into memory up front and that could be very slow.

GitUpKit/Interface/GIImageDiffView.m
```diff
@@ -0,0 +1,226 @@
...
+    if (newExtension.length) {
+      newPath = [newPath stringByAppendingPathExtension:newExtension];
+    }
+    if (![[NSFileManager defaultManager] fileExistsAtPath:newPath]) {
+      [self.repository exportBlobWithSHA1:_delta.newFile.SHA1 toPath:newPath error:&error];
+    }
+  } else {
+    newPath = [self.repository absolutePathForFile:_delta.canonicalPath];
+  }
+  _currentImageView.image = [[NSImage alloc] initWithContentsOfFile:newPath];
```
lucasderraugh, line 106: I'm a bit concerned about loading in the file this way in cases where the asset is rather large. For most cases we're probably not dealing with giant assets, but if you do this is going to grind to a halt. We should probably generate a thumbnail of a max size so we cap the load time. Another optimization we should probably have is to separate out the image loading and size calculations. The table view needs to know size (because we don't use automatic height adjustments yet, unfortunately) and the size calculation should be as fast as possible.

lucasderraugh (changes requested):
I don't think I can permit this at the moment as one of the main things we try to maintain with this app is performance. I'll test this out with large image assets today, but ultimately I think in its current implementation it would be a bit too slow. I'll try to make some suggestions as to what to change so we can get this feature in though.

An alternative would be to disable this by default and allow an opt in in preferences, but I feel that's more of a crutch than anything.

richardtop:
Looking forward to this feature! Looks great!

GitUpKit/Components/GIDiffContentsViewController.m
Stengo: Great point!
I switched it over to using [UTIs](https://developer.apple.com/documentation/uniformtypeidentifiers?language=objc), which should be very fast and still plenty reliable 🙂 (c4f52a4fbc19dd030341c6b220c7bf75d31b5216)

I'm not super familiar with memory management bridging for core foundation entities, though, so that part could definitely use a thorough look 😅 (the same goes for my other fix).

GitUpKit/Interface/GIImageDiffView.m
Stengo: Yeah, limiting the size in memory definitely sounds advisable!
I took care of that in 991f88908737795da132a25a4a2fef40e01c0364. For now I simply used a somewhat arbitrary maximum image dimension of 4000 pixels, because it was easy to do and seemed like a reasonable limit.
If you think we should change this value or even make it scale dynamically by screen size or something like that let me know 🙂

Stengo: With regards to separating out image loading and size calculations:
Does that mean you would want to make the cells a static height?

Right now we load the new images on `setDelta` and then query their dimensions during the sizing calculations to make sure they aren't scaled improperly and the aspect ratios make sense.
So the only way I could think of to avoid this reliance on in-memory images would be to set an arbitrary height and then cram the images into that space whenever they are ready 🤔

GitUpKit/Components/GIDiffContentsViewController.m
lucasderraugh: I think that should mostly be right, but to verify NSImageView can load it we probably should use `NSImage.imageTypes` to get the array of valid UTIs and then check conformance against those to see if we match any. I think `kUTTypeImage` type might be too generic.

Stengo: Sorted! bae1fc2d943165e92308ef3d57e030c3abd54e96

Stengo:
Hey @lucasderraugh!
Is there anything else I should be doing to help this PR along?
With the UTI identification in place (c4f52a4 / bae1fc2) and the thumbnails being generated (991f889) only the separation of image loading and size calculation would be left, but as mentioned in https://github.com/git-up/GitUp/pull/748#discussion_r646003595 I'm not sure how to go about that 🤔

lucasderraugh:
> Hey @lucasderraugh!
> Is there anything else I should be doing to help this PR along?
> With the UTI identification in place ([c4f52a4](https://github.com/git-up/GitUp/commit/c4f52a4fbc19dd030341c6b220c7bf75d31b5216) / [bae1fc2](https://github.com/git-up/GitUp/commit/bae1fc2d943165e92308ef3d57e030c3abd54e96)) and the thumbnails being generated ([991f889](https://github.com/git-up/GitUp/commit/991f88908737795da132a25a4a2fef40e01c0364)) only the separation of image loading and size calculation would be left, but as mentioned in [#748 (comment)](https://github.com/git-up/GitUp/pull/748#discussion_r646003595) I'm not sure how to go about that 🤔

Perhaps this StackOverflow post would answer this for you: https://stackoverflow.com/questions/1551300/get-size-of-image-without-loading-in-to-memory

I wish we didn't require size calculation up front, but that's a bigger API transition that is a bit beyond scope for this PR at least. If you can test with committing 30 or so large different images and see what the performance is like that would be great. I'll gladly test it out when the size logic is in.

Stengo:
Hello again @lucasderraugh!
I separated the image loading from the cell sizing as you suggested in https://github.com/git-up/GitUp/pull/748/commits/febb9950ceab945fc066d2fc0d3b61c397f07b28.
It essentially offloads the loading onto a different thread, but queries the file for it's size information right before doing so. It keeps the UI much more responsive for very large image files, but introduces a moment where a blank cell is shown. To mitigate that I introduced a spinner to indicate that things are still being processes 🙂 

Very large, compressed RAW camera files (like the ARW files created by Sony mirrorless cameras) still cause slight hiccups, but those are also present in Finder, so I doubt there's a way to work around them. Very large uncompressed RAW files (like big TIFF images) don't present this behavior.

I didn't experience any performance issues when dealing with a large number of high quality images (~40) beyond what the current live version of GitUp experiences (e.g. staging all files takes quite a few seconds).

Let me know what you think 🙂

lucasderraugh:
@Stengo Fantastic. I'll take a look when I get home to a laptop. We'll try to bundle up some changes for Monterey release.

GitUpKit/Interface/GIImageDiffView.m
```diff
@@ -0,0 +1,313 @@
...
+  }
+  _currentImageSize = [self imageSizeWithoutLoadingFromPath:newPath];
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+    NSImage* limitedSizeImage = [self generateLimitedSizeImageFromPath:newPath];
+    dispatch_async(dispatch_get_main_queue(), ^{
+      _currentImageView.image = limitedSizeImage;
+      [self setNeedsDisplay:true];
+    });
+  });
+}
```
lucasderraugh, line 125: Hi @Stengo. Testing this out it seems quite performant and looks amazing! My only question is around how images appear when they are deleted. The modified view shows the split view, the additional view shows the full image of the image added, but (at least for me) the delete image view shows a split with the left half being the deleted and the right being an empty image. Show the logic here be that if there is no new image that we don't draw it? I would expect to show the full deleted image.

lucasderraugh (changes requested):
Looks amazing and performant! One last question before we can get this in for me.

lucasderraugh (approved):

PR #772, opened 2021-10-12
GitUp. libgit2 updated. v1.3.0

lolgear:
This PR updates libgit2 to version v1.3.0.

Related PR in gitup libgit2 fork.
https://github.com/git-up/libgit2/pull/7

You need to run `sh ./update_xcode.sh` script to generate necessary file `features.h`


NikKovIos:
Merge please.

NikKovIos:
@lolgear fix conflicts please.

lucasderraugh:
I fixed them up. I can't ask @lolgear to fix changes all the time 😓.

lucasderraugh:
@lolgear Where is the `update_xcode.sh` script? When I don't see it as a file in this PR or the libgit2 PR.

lolgear:
Yes, it should use dirname $0.
Maybe a problem with invocation directory.
Do you try to run this script from libgit2 directory?

sh ./update_xcode.sh

lucasderraugh:
> Yes, it should use dirname $0. Maybe a problem with invocation directory. Do you try to run this script from libgit2 directory?
> 
> sh ./update_xcode.sh

@lolgear Okay, I needed to update my Xcode Command Line tools. The script is also called `update-xcode.sh` (dash not underscore). After running it though I still can't seem to run the GitUp project:
Screen Shot 2022-02-24 at 10 30 34 AM

lucasderraugh:
@lolgear Alright, so I figured out what I was doing wrong. It now runs, but I'm hitting this assertion in CGRepository:120:
```c
assert(git_libgit2_features() & GIT_FEATURE_SSH);
```

Any ideas how that could be resolved?

lucasdowlet:
Ya, unfortunately it's not being set correctly. My features.h file looks like this:
```
/* #undef GIT_SSH */
/* #undef GIT_SSH_MEMORY_CREDENTIALS */
```

lucasderraugh:
Ya, unfortunately it's not being set correctly. My features.h file looks like this:
```
/* #undef GIT_SSH */
/* #undef GIT_SSH_MEMORY_CREDENTIALS */
```

lucasderraugh:
@lolgear Is there a feature in Xcode 13.3 that makes the existing code work?

lucasderraugh:
@lolgear Did you test this under Xcode 13.3? If so, I can try it with Xcode 13.3 and push this out when it becomes public.

lolgear:
@lucasderraugh 
I didn't.
Well, let's wait for Xcode 13.3 release. I guess it will be released at the end of March.
I hope I will be available to do it at the end of March...

CodeCooler:
For what it's worth... 13.3 is out.

lucasderraugh:
Cool, I'll check it out this weekend. I still have to update my Xcode 😅

lolgear:
@lucasderraugh 

### Steps
I suggest following steps for merging PRs.

1. Merge PR https://github.com/git-up/libgit2/pull/8
2. Open PR with SPM plugin ( build tool ) which either wraps existing shell script OR reimplement it in Swift.3.

I'll implement second PR today.
Current PR still valid and can be merged without changes. ( If we want to or need to ).
It contains ( mostly ) changes in GitUpKit sources.

### Question about macro definitions differences in libgit2.
Also I want to ask libgit2 maintainers about current difference between macro definitions. It is very inconvenient that some definitions exist in features.h and some of them exist only as directives in preprocessor section in Xcode target.

### A problem with binaries targets and SPM.
The biggest problem here is figure out how to setup Package.swift, because it contains both third-party targets of libgit2 and binaries targets from other sources. Also this problem reveals another problem with plugin system in SPM - we have to define additional target with plugin and we have to put it somewhere.

### Improvements
Current setup is very clumsy and, actually, fragile.
SPM tool likes a lot of repositories, so... we can satisfy its wishes.

1. Add a repository with binaries targets. ( e.g. OurBinariesTargets )
2. Add OurBinariesTargets as dependency to Libgit2Origin target.
3. Add Libgit2Origin target to Xcode project as SPM Package and set branch and/or version anchor.

lucasderraugh:
@lolgear I thought your PR wasn't necessary if we were using Xcode 13.3? https://github.com/git-up/libgit2/pull/8
Is it still necessary and if so, why?

lolgear:
I tried Xcode 13.3 and plug-in support in SPM.

A short answer is as always - Apple releases a feature and hides a lot of details.

For example, I can't find an example in which plug-in can write to package directory.
However, Apple has a documentation for "PluginPermission" structure which has "writeToPackageDirectory" factory method. ( something that we need, yes ). But I can't find any occurrence of how to use it.

My PR contains a small improvement: force addition of definitions which we need. I thought that it can be done in SPM plug-in, but, as I mentioned above, we don't have any documented entity which allows to write to package directory.

If somebody has experience in SPM plugins ( especially in writing to package directory ), well, he/she can show how to do it.

lolgear:
Ok, I checked.
We can't ( right now ) use SPM _build_ plugin which writes to package directory. ( This plugin is "kind of" old "script phase" in Xcode ).
However, we can use SPM command line plugin ( if we want ).
This command line plugin may replace shell script. ( From this PR https://github.com/git-up/libgit2/pull/8 ).

lolgear:
@lucasderraugh Any updates?

lolgear:
Can we finally merge it?

lolgear:
@lucasderraugh Can you also open discussions ( enable them for this repository ), so, we can discuss ideas without opening issues.

lucasderraugh:
I'll check it again tonight. Note the past times I have checked the feature just hasn't worked with SPM and I'm very uncomfortable adding the risk of flaky SPM behavior. I'll do a deeper dive on the changes though to see where the rough edges are. Thanks for your patience.

lolgear:
@lucasderraugh 
Don't forget to keep it in sync with libgit2's gitup_updated branch.

Actually, I'm thinking about next steps towards better "management" of dependencies.
For example, it may be possible to completely remove submodules and do everything on top of SPM.
For that we need to remove XLFacility and to add SPM to GARawTracker ( if google stills supports these APIs ).
And last step is to move libgit2 dependencies ( libssh, libcrypto?, libssl? ) to libgit2 folder.
And after that it is easy to add `.binaryTargets`.

lucasderraugh:
@lolgear I removed GARawTracker in a recent commit. I'd like to get rid of the logging facility as a submodule as well but that needs a bit more thought.

lolgear:
@lucasderraugh 
Very good. Let's try to merge this PR first.
Then I would like to do two things.

1. Add proper SPM file to libgit2 to eliminate libgit2 submodule. This step will change project a bit, because it adds libssh and other libraries binaries to libgit2 repository ( I guess it is fine, because IF we do this, we can easily update dependencies and easily update libgit2 itself ).
2. Update libgit2 to current stable version ( 1.4.2 ). It will be easy, because we have `gitup_` prefixed functions.

lucasderraugh:
@lolgear Looks good mostly, but there are some concerns with the new warnings that are introduced:
Screen Shot 2022-05-24 at 9 24 03 PM

1. Do you know what the first one is referencing?
2. I think before I merge this I'll try to bring master up to 10.13 unless we don't have to define `GIT_USE_FUTIMENS`?

lolgear:
@lucasderraugh 

Invalid exclude is an entry in exclude list in Third-Party/Package.swift.
But second one, hm, it should work without futimens.
What version of Xcode/macOS do you use?

lolgear:
@lucasderraugh 
I updated exclude list and commit of submodule ( oh, submodules ).
Can you rerun update-xcode.sh again?

lolgear:
@lucasderraugh 
Can we merge this PR?
It is a blocker for further improvements ( as I described above ).

lucasderraugh:
@lolgear Sorry I was out sick for the past week, getting caught up on normal work but I'll visit this on the weekend.

lucasderraugh:
@lolgear Looks good from what I can tell. However, I want to fix our test suite before I merge this in to make sure I'm not missing anything. For some reason the xctest bundle isn't working on my end so I'll have to debug that a bit more tonight.

lolgear:
@lucasderraugh
Hope you are doing well.
I'm curious in futimens warnings appearance, because it should be fixed.
I asked libgit2 maintainers to include this fix before 1.3.0 release.
Could you tell your current setup?
macOS, Xcode, commits ( in GitUp / forked libgit2 ), contents of update_xcode.sh.

lolgear:
I checked this PR:

// update-xcode.sh
MACOSX_DEPLOYMENT_TARGET=10.10 cmake -S "$SCRIPT_DIRECTORY" -B "$SCRIPT_DIRECTORY/xcode" -G "Xcode"

and it's output is:

// sh update-xcode.sh
-- Looking for futimens
-- Looking for futimens - not found

and no warning at all...

lucasderraugh:
@lolgear Using Xcode 13.4, macOS 12.4. When I run it I get:
```
-- Looking for futimens
-- Looking for futimens - found
```

I'm on commit f2300f31af6050a2b9e9e56e808c357d8576c9e2 for this repo and b80dc34dd3b4abe59b5543efd849102ac75b7032 for libgit2 repo (our fork)

lolgear:
And what version of cmake?
I have cmake 3.22.3.
Do you run update-xcode.sh script without setting environment variables. MACOSX_DEPLOYMENT_TARGET should be 10.10.
It shouldn't find futimens at all, because deployment target is too low and futimens doesn't exist prior 10.13.

Could you check that
```
/// GitUpKit/Third-Party/libgit2/src/CMakeLists.txt 
/// contains
IF (HAVE_FUTIMENS)
  SET(GIT_USE_FUTIMENS 1)
ENDIF ()
```

and file `GitUpKit/Third-Party/libgit2/cmake/Findfutimens.txt` exists.

lucasderraugh:
I have `3.23.1` and CMakeLists have 
```
IF (HAVE_FUTIMENS)
	SET(GIT_USE_FUTIMENS 1)
ENDIF ()
```

lolgear:
I upgraded cmake to 3.23.2.

And yes

```
-- Looking for futimens
-- Looking for futimens - found
```

Yes, it is a bug in ( I don't know, ah ) cmake? ( again? )
Because cmake doesn't respect MACOSX_DEPLOYMENT_TARGET environment variable.

lolgear:
However, Apple will raise deployment targets to 10.13 in Xcode 14, so, it seems impossible to build products for 10.12 and less.

Not sure if it is necessary to fix this cmake issue.

https://developer.apple.com/documentation/Xcode-Release-Notes/xcode-14-release-notes

```
Building for deployment to OS releases older than macOS 10.13, iOS 11, tvOS 11, and watchOS 4 is no longer supported. (92834476)
```

lucasderraugh:
@lolgear I have most of the changes needed for the 10.12 bump. I'll try to get the build to 10.13 and then try this out again.

lolgear:
@lucasderraugh 
Can we merge this PR?

lucasderraugh:
Working on testing it out. It will be the next PR that goes into GitUp, I promise.

GitUpKit/Third-Party/Package.swift
```diff
@@ -0,0 +1,211 @@
...
+    }
+    
+    static func extraLibgit2CSettings() -> [CSetting] {
+        let fixedLibgit2CSettings = fixedExtraSSHDefines()
+        return fixedLibgit2CSettings
+    }
+}
+
+let package = Package(
+    name: "SwiftPackage",
```
lucasderraugh, line 70: @lolgear I found the source of my warnings with futimens. Turns out that swift packages defaults to 10.10 for its minimum target. I confirmed this with the build logs `-target arm64-apple-macos10.10`. After making the change I'm suggesting, it builds with `-target arm64-apple-macos10.13`. To bump this we need to define the following, let's but it below `name`:

```
platforms: [
    .macOS(.v10_13)
],
```

lucasderraugh (changes requested):
Requesting above changes.

GitUpKit/Third-Party/Package.swift
lolgear: Well, all warnings are coming from broken CMake. If you use CMake 2.22, it will take into account environment variable in script and correctly cut out futimens for libgit2.
But recent version of CMake uses a different mechanism and ignores current environment variable.

lucasderraugh: Well I don't want to rely on a particular version cmake. Let's define the platform and we'll always get what we want. I've already bumped to 10.13.

lucasderraugh: Also...I don't think the warnings (for this issue) are on CMake's end for me, it seems like it compiles correctly with that value looking at the CMake output logs. It's when compiling the `Libgit2Origin` target in Xcode that this warning surfaces:

```
CompileC /Users/lucasderraugh/Library/Developer/Xcode/DerivedData/GitUp-bqeuvsqkhipodmffbhvhltdikopc/Build/Intermediates.noindex/SwiftPackage.build/Debug/Libgit2Origin.build/Objects-normal-asan/arm64/xprepare.o /Users/lucasderraugh/Developer/GitUp/GitUpKit/Third-Party/libgit2/src/xdiff/xprepare.c normal arm64 c com.apple.compilers.llvm.clang.1_0.compiler (in target 'Libgit2Origin' from project 'SwiftPackage')
    cd /Users/lucasderraugh/Developer/GitUp/GitUpKit/Third-Party
    export LANG\=en_US.US-ASCII
    /Applications/Xcode-13.4.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -x c -target arm64-apple-macos10.10
```

lolgear: It works with previous CMake and libgit2 maintainers added futiments function to search for futimens function, because CMake can't find it.
It works for previous CMake and now it doesn't :)
CMake decided to change something and here we are :)
I'll update both constant in script and platform version.

lolgear:
@lucasderraugh I added supported platforms.
I guess that you can raise also deployment target version in update-xcode.sh script (in libgit2 repository), but it doesn't work now so, not a big trouble.

GitUpKit/Third-Party/Package.swift
lucasderraugh: Thank you I greatly appreciate your work on this! I'm out tomorrow but will try to merge this in as long as the tests still work on the Mac side (the iOS ones I think are already broken though I should probably fix them)

lucasderraugh (changes requested):
Unfortunately, the `GitUpKit (macOS)` tests don't compile, and even after fixing that, they don't pass. I'll see if I can figure out what's causing the failures, but we can't merge this in until the macOS tests are passing.

lolgear:
It is a hard decision, but... I suggest to test it in pre-release.
Actually, this project requires a lot of time for modernization. ( remove MRC, move to SPM, cleanup tests/headers etc. )

GitUpKit won't pass tests, because so many changes happened in past years and libgit2 shows a lot of errors in tests.

lucasderraugh:
The tests are there to prevent this exact issue. We need to validate that we haven't broken existing functionality or if we are understand why. I've reviewed some already but it will take some more time on my end.

lucasderraugh:
Closing as I had other changes I needed to make before merging in and can't push to this PR. Thanks again for the help.

lolgear:
@lucasderraugh 
I suggested to use a different branch ( dev? ) for this PR.
Instead you decided to fix tests.
Also I believe that this PR (#852) is the most valuable.

Issue #788, opened 2021-12-08
Unstaged diff for image files is broken


Stengo:
It seems like 8147b529e1858c1d9c8dd6c7ca547e9446eeb6c5 broke the visualization of unstaged image files.

The previous fallback behavior to the canonicalPath has been replaced by nilling out the property [here](https://github.com/git-up/GitUp/blob/master/GitUpKit/Interface/GIImageDiffView.m#L123).

FYI @lucasderraugh 

![Screenshot 2021-12-08 at 09 25 44](https://user-images.githubusercontent.com/12018882/145178684-d8781080-f5fe-4ba0-a07b-33cba7f5a7fc.png)

lucasderraugh:
@Stengo Should be fixed in 1.3.1 on continuous channel.

PR #789, opened 2021-12-08
Fallback to canonical path for unstaged images

Stengo:
Fixes https://github.com/git-up/GitUp/issues/788

This rolls back to the original behavior of falling back to the canonical path for unstaged image files that do not have a SHA1 value yet.

### DIFF
--- a/GitUpKit/Interface/GIImageDiffView.m
+++ b/GitUpKit/Interface/GIImageDiffView.m
@@ -111,17 +111,17 @@
     if (![[NSFileManager defaultManager] fileExistsAtPath:newPath]) {
       [self.repository exportBlobWithSHA1:_delta.newFile.SHA1 toPath:newPath error:&error];
     }
-    _currentImageSize = [self imageSizeWithoutLoadingFromPath:newPath];
-    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-      NSImage* limitedSizeImage = [self generateLimitedSizeImageFromPath:newPath];
-      dispatch_async(dispatch_get_main_queue(), ^{
-        _currentImageView.image = limitedSizeImage;
-        [self setNeedsDisplay:true];
-      });
-    });
   } else {
-    _currentImageView.image = nil;
+    newPath = [self.repository absolutePathForFile:_delta.canonicalPath];
   }
+  _currentImageSize = [self imageSizeWithoutLoadingFromPath:newPath];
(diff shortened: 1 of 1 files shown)


lucasderraugh:
Hmm, not sure why I would have changed that logic, I feel like there was another reason I had in mind, perhaps it was the deletion case where I was trying to only see the old image and not have the split view.

lucasderraugh (changes requested):
Could we replace the `updateCurrentImage` call with the following:
```obj-c
- (void)updateCurrentImage {
  NSError* error;
  NSString* newPath;
  if (_delta.newFile.SHA1 != nil) {
    newPath = [GILaunchServicesLocator.diffTemporaryDirectoryPath stringByAppendingPathComponent:_delta.newFile.SHA1];
    NSString* newExtension = _delta.newFile.path.pathExtension;
    if (newExtension.length) {
      newPath = [newPath stringByAppendingPathExtension:newExtension];
    }
    if (![[NSFileManager defaultManager] fileExistsAtPath:newPath]) {
      [self.repository exportBlobWithSHA1:_delta.newFile.SHA1 toPath:newPath error:&error];
    }
  } else if (_delta.canonicalPath && _delta.oldFile.SHA1 == nil) {
    newPath = [self.repository absolutePathForFile:_delta.canonicalPath];
  }
  if (newPath) {
    _currentImageSize = [self imageSizeWithoutLoadingFromPath:newPath];
    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
      NSImage* limitedSizeImage = [self generateLimitedSizeImageFromPath:newPath];
      dispatch_async(dispatch_get_main_queue(), ^{
        _currentImageView.image = limitedSizeImage;
        [self setNeedsDisplay:true];
      });
    });
  } else {
    _currentImageView.image = nil;
  }
}
```

This way we'll show unstaged images and the deleted assets won't show with the diff view (we only want to see the full asset that is being deleted).

Stengo:
I just tried out your code and I'm still seeing the deleted asset in the diff view, the behavior appears identical to the code in this PR 🤔 
But I might be misunderstanding the exact scenario 😅

lucasderraugh:
Sorry, that was an incorrect explanation. The only difference in the code paths would be that in your code we would be generating 2 images in the deleted case (the old and "new" which is the same asset). My code suggestion would be skipping over creating an image for the current image when we only need to be drawing the old image.

I'll push my changes to master and get a build for continuous later tonight. Thank you for bringing this to my attention and I'm sorry I broke your excellent work.

lucasderraugh:
Going to close this PR out as I've applied the fix directly to master. Thanks again.

PR #795, opened 2021-12-17
Fix #792 Missing Image Comparison

### LINKED ISSUE #792: Modified image shows up as deleted
This appears to be a follow-up issue to https://github.com/git-up/GitUp/issues/788.

Right now, when adding a file, staging it and then modifying it, the image preview is showing it as having been deleted instead of modified (there is no comparison available).
The original code (which was proposed as a roll back in https://github.com/git-up/GitUp/pull/789 ) seems to fix this issue along with the previous one.

FYI @lucasderraugh 

Screenshot 2021-12-15 at 00 26 50

Stengo:
Fixes https://github.com/git-up/GitUp/issues/792 by reverting to the original solution.


lucasderraugh:
I'll push an update later today for the continuous build.

PR #807, opened 2022-03-08
Add horizontal width constraint for the toolbar navigation control

bogdanbolchis:
On macOS 12.2.1 (Monterey) the three button segmented control is clipped
on the right side, making the "Stashes View" segment shorter than normal.

This change adds a width constraint on the segmented controls, which
fixes the problem on 12.2.1.

### DIFF
--- a/GitUp/Application/Base.lproj/Document.xib
+++ b/GitUp/Application/Base.lproj/Document.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="17506" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="19162" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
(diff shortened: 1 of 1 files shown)


bogdanbolchis:
Hi all!

This bugfix PR should help with a visual issue on macOS 12 (Monterey). Here is a screenshot of the issue:

Screenshot 2022-03-08 at 14 45 03

This should be tested on previous macOS versions. Unfortunately I don't have access to macOS < 12.

lucasderraugh:
Thanks for looking into this! It bothers me that this needs a constraint at all considering it resolves itself if you readjust the window size. In theory the intrinsic content size should be enough with the given constraints, though clearly that isn't the case. I'll try it out this weekend and I'll take this change if I don't see an obvious resolution.

JokerQyou:
> ...it resolves itself if you readjust the window size. ...

It's strange that if I manually resize the window size (by dragging at the window border) the issue resolvers itself; but when using window manager utilities (like Magnet) to set window to half the screen width it would not.

lucasderraugh:
@bogdanbolchis I ended up going with just fixing the widths of the segmented control 628f48bdee34466bcecaa49cc3a313e47086acf8. I appreciate you doing this so long ago and I apologies for not giving quicker turnaround.

bogdanbolchis:
Thanks! I also noticed today for the first time that resizing the window updates the width and fixes it. If i knew this maybe i would have skipped a PR. No idea why the intrinsic size isn’t honored, though i did see a bunch of AutoLayout warnings in the same xib. Also i’m not sure if that’s influencing layout.

danielsaadaudajr:
Great, thanks.

PR #845, opened 2022-06-23
Change Commit shortcut to ⌘+Return 

danshevluk:
Hey! 
I've been using GitUp for a long time for all kinds of tasks and thought of a small improvement that could improve most users' workflow.

This PR changes the shortcut for creating a commit from the `AdvancedCommitView` to `Cmd+Return` instead of `Opt+Return`.  Usage of a `Cmd+Return` is more common in modern macOS apps, so it makes sense to update option usage to a more familiar key binding.

Similar shortcuts could be found in other Git GUI like `Fork`, `Tower`, or `GitKraken`. 
[Keycombiner](https://keycombiner.com) provides a convenient way to compare all three:
- https://keycombiner.com/collections/fork/
- https://keycombiner.com/collections/tower/
- https://keycombiner.com/collections/gitkraken/

To complete this task it's needed to update [wiki page with sourtcuts](https://github.com/git-up/GitUp/wiki/Using-GitUp-Advanced-Commit-View#keyboard-shortcuts), but it's impossible to do in a PR, so please help me with that.

Looking forward to your feedback!

### DIFF
--- a/GitUpKit/Views/Base.lproj/GICommitRewriterViewController.xib
+++ b/GitUpKit/Views/Base.lproj/GICommitRewriterViewController.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="14490.70" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="20037" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="14490.70"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="20037"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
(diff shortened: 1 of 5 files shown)


lucasderraugh:
Thanks for the contribution! While I agree that it is the more natural shortcut, I'm also a bit concerned with dropping the old way considering the whole app uses option as the main modifier (for better or worse). Perhaps there is a way to preserve option enter in addition to adding command. I'll have to think about it a bit. Anyone else is willing to comment how they feel as well.

peteschaffner:
I've been maintaining a personal fork of GitUp and have made this same change (in fact I didn't realize the option modifier binding was in place as it isn't a shortcut I've encountered before in situations with a textfield + primary button).

As @lucasderraugh mentioned, it is the main modifier throughout, so if the change were to be made, it should really be done everywhere. (Look at all the view controller `xib`s that are presented as modal views via `runModalView:withInitialFirstResponder:completionHandler:`.)

I also agree that if opt-return could be maintained, it would be great for those who have become accustomed to it (although I'm not sure how to go about that).

danshevluk:
Thanks for the feedback! I'll try to find a way to support both shortcuts and will check for other similar actions 👍

danshevluk:
OK, so after some research, it appears that there are a lot of places with the same behavior. Currently those are:

-> `GIAdvancedCommitViewController.xib`
- [x] **Commit**

-> `GICommitRewriterViewController.xib`
- [ ] **Rewrite Commit**

-> `GICommitSplitterViewController.xib`
- [ ] **Continue…**
- [ ] **Split Commit**

-> `GISimpleCommitViewController.xib`
- [ ] **Commit All Changes**

I tried to find solutions to support both `keyEquivalents`, but could not find any easy options. There is a way to add an additional `menuItem` with the same actions but different key binding. This could work for some actions, but not as a whole-project solution.

If you agree that it's an important addition to GitUp, then there are two ways that we achieveachive that:
- Just change the shortcut in all mentioned places to `cmd+return`. It's easy to do, but could break old shortcuts for lots of users ⚠️ 
- Move `keyEquivalents` setup to `.m` files from `.xib` and switch between option/cmd based no user preferences. For all new users it could be set to `cmd`, and to keep old users happy they will be able to opt-in in preferences, but not automatically.

Between these two options, I irrationally tend more to make a breaking change. The original way is strange and hard to discover without specifically looking into documentation.

How do you feel about these options?

peteschaffner:
I would also just make the breaking change in this particular situation.

(I suppose one way of teaching about the change would be to throw an alert when the old command is used, explaining the new modifier.)

lucasderraugh:
Alright, let's go with the breaking change of using command for the commit editors everywhere in the app. I'll play around with it before pushing an update to see if there needs to be any type of coaching.

danshevluk:
@lucasderraugh done! I updated all mentioned shortcuts to cmd+return as discussed. Please let me know if you have any further suggestions ✅

braver:
I'd just like to chime in to say this is pretty massive. I use several tools to work with Git repos and my fingers get tangled up because Gitup is different from everything else, especially for submitting commit messages.

🎉 👍🏻

lucasderraugh:
Thanks @danshevluk. I want to test out some other libgit changes that I've needed to get in for some time before we put this in, but I'll pick this up after that.

lucasderraugh:
Sorry for the long delay on this. Just pushed 1.3.3. to the continuous channel. I would like to work out some kind of alert dialog that shows once for users informing them of this change as it is a muscle memory changer for sure.

I'll try to think of some simple UI and how it can be tracked, but I'm just busy with lots of other things unfortunately so it might not be right away.

GitUpKit/Views/Base.lproj/GICommitRewriterViewController.xib
```diff
@@ -41,34 +41,34 @@
                                 <color key="backgroundColor" name="textBackgroundColor" catalog="System" colorSpace="catalog"/>
                             </textFieldCell>
                         </textField>
-                        <textField horizontalHuggingPriority="251" verticalHuggingPriority="750" id="hZw-Mr-R8r">
+                        <textField horizontalHuggingPriority="251" verticalHuggingPriority="750" fixedFrame="YES" translatesAutoresizingMaskIntoConstraints="NO" id="hZw-Mr-R8r">
                             <rect key="frame" x="135" y="8" width="847" height="16"/>
                             <autoresizingMask key="autoresizingMask" widthSizable="YES" flexibleMinY="YES"/>
                             <textFieldCell key="cell" controlSize="small" lineBreakMode="truncatingTail" allowsUndo="NO" sendsActionOnEndEditing="YES" alignment="left" title="&lt;TITLE&gt;" id="rrE-b8-3j3">
-                                <font key="font" metaFont="controlContent"/>
+                                <font key="font" metaFont="cellTitle"/>
```
NikKovIos, line 48: nit: is it right change? in other places there is controlContent font.

richardtop:
Very interested in this change. Would feel more intuitive to use compared to the default.

prostolyubo:
Same, cmd is much better modifier.

danlucraft:
Thank you for this! For years I've been hitting Cmd+Return in gitup to commit and every time hearing the bonk noise 😄

richardtop:
Now I find it difficult to switch back...

ilyasotkov:
Surely this change will feel more intuitive for new users, but for people who've been using GitUp daily for years, this feels like banging your head against a wall every time you're trying to make a commit because there's some serious muscle memory already built up.

Would be great to be able to have to option to go back to `Opt+Return`.

tockrock:
Thanks for this! I appreciate the change.

PR #921, opened 2023-03-11
Add key navigation between index and workdir files

fo2rist:
Navigate to index files controller when Down button pressed while on the last file of the workdir list
Navigate to workdir files controller when Up button pressed while on the first files of the index list

**Context**: I'm using keyboard navigation for most of the actions in GitUp but there were no way to quickly navigate b/w staged and un-staged files, and while it's possible with Tab/Shift+Tab it breaks natural up/down navigation that's available in simple mode. This change adds that ability to standard mode as well.

**Testing and considerations**:
- I checked that simple mode still works the same as before
- I added checks to make sure there is not confusing behavior when Arrow key is pressed with a modifier or when multiple files selected to avoid losing selection
- I added checks to prevent navigation to an empty list
- No changes to the documentation page is needed since it simply mentions `Arrow up & Arrow down to select a different file` which is still true

### DIFF
--- a/GitUpKit/Interface/GIConstants.h
+++ b/GitUpKit/Interface/GIConstants.h
@@ -31,4 +31,4 @@ typedef NS_ENUM(unsigned short, GIKeyCode) {
   kGIKeyCode_Delete = 0x33
 };
 
-static const long kGIKeyModifiersAll = NSEventModifierFlagShift | NSEventModifierFlagControl | NSEventModifierFlagCommand | NSEventModifierFlagOption;
+static const NSEventModifierFlags kGIKeyModifiersAll = NSEventModifierFlagShift | NSEventModifierFlagControl | NSEventModifierFlagCommand | NSEventModifierFlagOption;
--- a/GitUpKit/Interface/GIConstants.h
+++ b/GitUpKit/Interface/GIConstants.h
@@ -13,6 +13,7 @@
 //  You should have received a copy of the GNU General Public License
 //  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+#import <AppKit/NSEvent.h>
 #import <Foundation/Foundation.h>
 
 typedef NS_ENUM(unsigned short, GIKeyCode) {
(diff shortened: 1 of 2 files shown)


fo2rist:
Would like to discuss a possibility to add arrow navigation between individual lines in the diff view.
That would allow me to do complete change/review/commit cycle with keyboard.

lucasderraugh:
Thanks for writing this up. I'll take some time tonight or tomorrow to review what's being changed here.

GitUpKit/Interface/GIConstants.h
```diff
@@ -29,3 +30,5 @@ typedef NS_ENUM(unsigned short, GIKeyCode) {
   kGIKeyCode_Return = 0x24,
   kGIKeyCode_Delete = 0x33
 };
+
+static const long kGIKeyModifiersAll = NSEventModifierFlagShift | NSEventModifierFlagControl | NSEventModifierFlagCommand | NSEventModifierFlagOption;
```
lucasderraugh, line 34:
```suggestion
static const NSEventModifierFlags kGIKeyModifiersAll = NSEventModifierFlagShift | NSEventModifierFlagControl | NSEventModifierFlagCommand | NSEventModifierFlagOption;
```

GitUpKit/Views/GIAdvancedCommitViewController.m
```diff
@@ -319,6 +320,27 @@ - (BOOL)diffFilesViewController:(GIDiffFilesViewController*)controller handleKey
     }
   }
 
+  // Navigate beteween stated and unstaged files list by up/down arrows
+  long modifiers = event.modifierFlags & kGIKeyModifiersAll;
```
lucasderraugh, line 324:
```suggestion
  NSEventModifierFlags modifiers = event.modifierFlags & kGIKeyModifiersAll;
```

lucasderraugh (changes requested):
Apologies for the delay in reviewing this. Looks good to me and welcome functionality.
Small nits just to change the type to match the `modifierFlags` property.
I was also thinking if there was an easier way to just say there were no modifiers, but I guess it includes caps lock and things like that, so this looks appropriate.

Other than that, it's ready to merge.

GitUpKit/Interface/GIConstants.h
fo2rist: Done

lucasderraugh (approved):

Issue #216, opened 2016-09-13
Increase the contrast of the Dock Icon


stevendaniels:
Thanks for gitup. It's a great app.

The contrast of the icon is pretty low.  When switching between applications (CMD+Tab),  it almost looks like a white square.

sirnacnud:
I'm not for sure what the icon is supposed to represent or signify, but it doesn't stand out when compared to other apps. It looks like some kind of default app icon.

muescha:
i would also vote for a better icon. some icon ideas are in #396

sesam:
Yep, the dock icon is completely impossible to see -- just a white blur in most lighting conditions.
#good-first-issue

PR #945, opened 2023-09-02
Application icon

davidfmiller:
A proposed fix for issues https://github.com/git-up/GitUp/issues/216


lucasderraugh:
Awesome icons! I'll chat with Pierre about getting access to the GitUp website so I can change the contents there. Would want to be able to do that before committing to a new icon just yet.

As for the icon I think it's much more representative of "GitUp". I might play around with different colors (perhaps a blue instead of red) but other than that I think it looks wonderful. Thanks for putting in the effort.

mtrudel:
This icon looks fantastic! One of those designs that is so simple and obvious in hindsight. Well done!

As someone who's been bit by this countless times with app icons, the source SVGs should 100% be in source control! 

Also in terms of variants (when it gets to the preferences option for them), some less saturated options would be good to have as well, if just to appease folks (like me) who are so used to getup being a white icon.

davidfmiller:
I added the icon (in pink) in SVG format to the branch, not sure how to handle the different variations (or which variations I should include)?

davidfmiller:
Excellent! Let me know if you need a hand with anything, I do the webs as well.

lucasderraugh:
I think I'm going to just take the icon as is. I like the base color here and maybe I'll add the ability to choose later on, but for now I think this is a nice improvement.

mdznr:
Just updated to GitUp 1.4.x from 1.3.5. This icon is far too bright and draws way too much attention to itself in the dock. The app icon‘s shadow doesn’t match other icons in the dock either.

lucasderraugh:
I'm open to proposed changes. We can also give the option to show other icons, but ultimately I think it's still better than the original in describing what the app does.

imawizrd:
The color red is often associated with error statuses or notifications. In macOS, icon badge labels also use a red color. This may be why @mdznr mentioned that it stands out too much on the dock and is distracting. Using a widely recognized accent color for different purposes can be confusing.

mattgallagher:
The pink was really intense on my sRGB LG screen (it was fine on my MacBook's P3 display but that's not where my Dock usually sits). I like the design and I also don't want everything to just be "blue" but I ended up dialling it back to something a little more neutral.

Screenshot 2024-04-08 at 3 36 09 PM

lucasderraugh:
I always love when I finally send things to the stable channel 😅. @mattgallagher I think this seems to be the sentiment in general is that the color itself is perhaps just too bright. I think I'll do almost what you did here which is to tone down the red a bit and make it a bit closer to the red branch line in the app.

perifer:
Love that this app is still maintained. Thanks for all the hard work you put into it. I can see why this seems like a relative big change... Maybe an option could be to add a setting for the icon so one can revert to the old one?

richardtop:
I liked the previous one more :(

Cykelero:
As a counterpoint, I'd like to note that I really like the new icon, including its strong color. The previous one really did lack definition, to me.

codymikol:
I'm not at all complaining about the change btw, always appreciate anyone who take the time to contribute to FOSS ❤️❤️❤️

indeyets:
@lucasderraugh website still needs to be updated with a new icon

chipotle:
> The color red is often associated with error statuses

Come on, Apple Music isn't THAT bad!

red-icons

(Slightly more seriously, I think red is just fine, but the option to switch between icon variants discussed earlier in the comment thread seems like it's probably a good idea. Maybe include the original icon for people who ~~hate fun~~ would rather not change.)

davidfmiller:
> @lucasderraugh website still needs to be updated with a new icon

FWIW if a hand is required in updating / maintaining the GitUp website I'd be happy to lend a hand with that… I would be sure to make it very bland and milquetoast in hopes of not ruffling people’s feathers 😊

davidfmiller:
@lucasderraugh will do! I’ll double-check the drop shadow used on the round rect, too… though I’m 99% sure I used the template that Apple provides for app icons. 

Screenshot 2024-04-24 at 09 54 52

🫠

imawizrd:
> > The color red is often associated with error statuses
> 
> Come on, Apple Music isn't THAT bad!
>
> (Slightly more seriously, I think red is just fine, but the option to switch between icon variants discussed earlier in the comment thread seems like it's probably a good idea. Maybe include the original icon for people who ~hate fun~ would rather not change.)

Are you ok? I did not mean to stir up strong feelings about this, I was just expressing my opinion.

Not sure how a music player has anything to do with a git GUI by the way. I mean its fine if they want to be more flashy.

I like having the option to change the app icon in the app itself, but note that macos already gives that option to users, I think you can drag an image file to the information details window in Finder for any app.

NikKovIos:
After a year with the white GitUp icon, I really want to continue using it. However, now I am using a workaround from this issue with every GitHub update, which is, of course, inconvenient. Therefore, I would like to be able to select the icon directly from the app.

https://github.com/git-up/GitUp/issues/984

codymikol:
Try this

* pin gitup to the taskbar
* physically print out the previous icon
* tape it over the gitup icon.

You can actually keep a drawer full of these and change them throughout the day.

I have one for gitup at coffee time, while I'm traveling, when I'm on vacation etc.

This can actually work cross device as well and continues to function while the computer is turned off so you can still admire that crisp original logo during a power outage.

Cykelero:
More seriously, you might enjoy [Pictogram](https://pictogramapp.com/) as a more convenient workaround. It lets you change the icon of any app, and automatically reapplies the icon after app updates.

PR #965, opened 2024-01-30
Trimming whitespaces from url while cloning (Fixes #952)

### LINKED ISSUE #952: Trim space characters in the clone bar
When copying a url from somewhere else it's often possible to accidentally copy some space characters in front or after the actual URL.

For example, instead of `git@github.com:git-up/GitUp.git` this will be copied: ` git@github.com:git-up/GitUp.git `

As the result, GitUp shows this alert:
Screenshot 2023-10-18 at 19 27 23
Would be great to simply ignore these white space characters and continue to clone the actual URL.

prostolyubo:
Added in CloneWindowController.m a trimming operation before passing contents of a text field into GitUpKit
Fixes git-up/GitUp#952

### DIFF
--- a/GitUp/Application/CloneWindowController.m
+++ b/GitUp/Application/CloneWindowController.m
@@ -68,7 +68,7 @@
     [self beforeRunInModal];
   }
   if ([NSApp runModalForWindow:self.window] && self.urlExists) {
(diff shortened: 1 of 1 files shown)


richardtop:
Thanks

lucasderraugh (approved):
Awesome! Thanks for picking that up.

Issue #957, opened 2023-12-07
Explain blue dot on "pull" button


JoeStrout:
I've been using GitUp for a couple weeks, and I can't for the life of me understand why sometimes the "pull" button (near bottom-left of the map view) has a blue dot in it.  I've got a blue dot right now, but when I click the button it just tells me 'the branch "main" is already up-to-date with its upstream!'  I have no local changes, either.  But sometimes the blue dot goes away, so I know it's not a permanent feature of the button.

This blue dot is not explained by the hover tip, nor is it mentioned in the [documentation](https://github.com/git-up/GitUp/wiki/Using-GitUp-Map-View).

Concrete suggestions:

- When you apply (or remove) the blue dot, update the help tag.
- Add an explanation to the "[using GitUp map view](https://github.com/git-up/GitUp/wiki/Using-GitUp-Map-View)" page of the docs.

PR #970, opened 2024-01-31
Update tooltip of a pull button when it changes (#957)

prostolyubo:
The change in pull button when the tip is behind is also reflected in an updated tooltip. Resolves git-up/GitUp#957

Also, the following line should be added to the [wiki page](https://github.com/git-up/GitUp/wiki/Using-GitUp-Map-View)'s **Tips** section:
- When the local tip is behind the origin a blue dot is shown next to a pull button to indicate new changes on the remote.


lucasderraugh (commented):
Looks good, thanks!

lucasderraugh (approved):

Issue #897, opened 2022-12-18
Snapshot Preview titlebar item is selected when it now showing


japanese-goblinn:
This how it looks 
<img width="248" alt="image" src="https://user-images.githubusercontent.com/33631005/208271902-60034843-fdd0-483f-a6d3-3fb0b708f567.png">

This is how it should look
<img width="257" alt="image" src="https://user-images.githubusercontent.com/33631005/208271912-041a55cd-7b2d-4809-b39a-6a7d443decd2.png">

lucasderraugh:
How did you reproduce this, as I don't see this. Was it a one off? What version of macOS are you running?

japanese-goblinn:
@lucasderraugh macOS 13.0.1, Its happening when I press the Snapshot Preview button and then close it with `esc`

prostolyubo:
Yup, happens even if you close it normally by pressing the button again

PR #971, opened 2024-02-01
Correcting snapshot toggle state when navigating using escape button

prostolyubo:
As the Snapshot view button is not part of the navigation group its state is not cleared when the view is switched programatically. Therefore another correction is needed to set the toggle state to off if exited using escape button.
Fixes git-up/GitUp#897

### DIFF
--- a/GitUp/Application/Document.m
+++ b/GitUp/Application/Document.m
@@ -1279,6 +1279,8 @@ static NSString* _StringFromRepositoryState(GCRepositoryState state) {
           handled = YES;
         } else if (_snapshotsView.superview) {
           [self toggleSnapshots:nil];
+          // Correct button toggle state if toggled by esc
+          [(NSButton *)_snapshotsItem.view setState:NSOffState];
           handled = YES;
         } else if (_reflogView.superview) {
           [self toggleReflog:nil];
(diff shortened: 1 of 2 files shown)


GitUp/Application/Document.m
```diff
@@ -1279,6 +1279,8 @@ - (BOOL)windowController:(GIWindowController*)controller handleKeyDown:(NSEvent*
           handled = YES;
         } else if (_snapshotsView.superview) {
           [self toggleSnapshots:nil];
+          // Correct button toggle state if toggled by esc
+          [(NSButton *)_snapshotsItem.view setState:NSOffState];
```
lucasderraugh, line 1283: Thanks for fixing this! Could I ask that we either do an if conditional on the `_snapshotsItem.view` to `NSButton` conversion or perhaps better yet we make it a computed property somewhere called `snapshotButton` that we safely cast to `NSButton`. I just prefer not to crash in a case where this view was later wrapped or changed in the future.

lucasderraugh (changes requested):

GitUp/Application/Document.m
prostolyubo: You were more than right, as this would not work if f.e. later a shortcut would be used to open/close tha shortcuts view. I now safely cast (and inform future devs if the layout ever changes) and just synchronize the state of the view with the model so it will display correct state regardles of what was used to open/cole the pane.
I also added a tooltip as I noticed it was missing.

lucasderraugh (approved):

PR #979, opened 2024-02-20
Update copyright year in Info.plist

chrisbrandow:
I personally wondered if I had the most recent version of Gitup running on my computer because the About window listed the copyright only going until 2019. Since the app is being regularly updated, the copyright date should reflect that.


lucasderraugh (approved):
Now I just have to actually push out a release 😅

Issue #161, opened 2016-06-06
Unable to Push/Pull with Error: authentication required by no callback set


bjg222:
I have a simple git repo that I cloned in the command line, then opened via GitUp.  I can push/pull via command line just fine, but when I try to the buttons in the lower right hand corner of GitUp, I get the error seen in the attached image.  My server is on my local work network and does require authentication via SSH, for which I have a trusted key set up for passwordless SSH.  I have been unable to find a setting that relates to push/pull that the error message could be referring to.

<img width="1006" alt="screen shot 2016-06-06 at 6 13 04 pm" src="https://cloud.githubusercontent.com/assets/11563394/15839714/edd24ace-2c12-11e6-81c0-4b262a1156e3.png">

swisspol:
Make sure your Git remote URL uses a standard format and see https://github.com/git-up/GitUp/wiki.

bjg222:
I've reviewed the known issues before posting, but I'm not sure how that helps me?  The page doesn't refer to the remote URL format at all.  Also, I do not control the remote server, but just use it.  The URL I am provided is this: server.company.com:/repo/tools

libmaru:
@bjg222 You do have set up password-less authentication for SSH, but not for GitUp.

According to https://github.com/git-up/GitUp/wiki:
> * SSH config files (typically located at ~/.ssh/config) are ignored by GitUp since it uses libssh2 and not the ssh tool for SSH remote operations.
>   * GitUp first attempts to find SSH private keys using the ssh-agent
>   * It then tries all private key files in ~/.ssh/

GitUp found ssh-agent instance for your login session and ask for available private keys, which does not include your non-default SSH identity.
To solve your problem, `ssh-add -K <path_to_your_key>`.

wprater:
I have a key added and Im still getting this error.  any ideas?

LJeub:
I was having the same issue and running `git fetch` on the repo from the terminal resolved it for me. No idea why that worked?

luongvo2009:
I believe a solution for this issue was discussed here https://github.com/git-up/GitUp/issues/35

xtazz:
Still having this issue cause I have IdentityAgent configured in ssh config, but it's ignored by GitUp. Can I somehow config GitUp to make it work?

SenHeng:
This is probably an old issue but it's the first result that pops up on google so for future google arrivers here is the issue.

[...]

Issue #35, opened 2016-03-02
Improve error message when SSH key is locked


swisspol:
When an SSH key is password protected and not added to the ssh agent, the error gitup produces is:

    Failed connecting to "origin" remote: authentication required but no callback set

Gitup should either produce a sane error message explaining that the SSH key cannot be unlocked or (better) prompt for the SSH key password (ideally add it to ssh agent).

davisonio:
How can I work around this? Pushing from command line works but I can't push using gitup.

andreyz:
I think the diagnosis of this error is incorrect. ssh agent and up and running and keychain is unlocked. Gitup still refuses to do a remote push.

andreyz:
Any hints on what's going @swisspol ?

swisspol:
No idea, sorry. It's not something I have observed myself so someone would need to build GitUp locally and trace in the code what is happening.

TheSpyder:
I'm not sure what configuration causes this, but on my machine with the problem SSH asks for the key password in terminal instead of using the normal keychain dialog. It also doesn't seem to store the passwords in any place keychain access can find them, despite `ssh -vv` showing it is getting the password from keychain. I didn't realise this had happened until I tried to use GitUp and hit this issue.

I wonder if it's related to iCloud keychain?

I can't compile GitUp on this machine because I'm running 10.12 beta, on my 10.11 test machine where I was able to compile it SSH asks for key passwords the normal way and GitUp is fine. I did notice however that if I haven't unlocked a key at the console on that machine I get the same "no callback set" error, with this in the log:

`[ERROR    ]> libgit2 error (-1): authentication required but no callback set`

swisspol:
@TheSpyder why is GitUp not compiling on 10.12?

TheSpyder:
`ld: library not found for -lcrypto`

It might just be the Xcode beta, but the macOS 10.12 SDK is missing library files like libcrypto and libz that I can see are there in the 10.11 SDK. I found them in /usr/lib on my system but when I tried to make the build use those it still gives me the error.

swisspol:
See #210 for the 10.12 compiling issue.

TheSpyder:
yeah, I didn't mean to turn this into a 10.12 compiler issue.

[...]

PR #987, opened 2024-04-13
Check if `SSH_AUTH_SOCK` env var is set before trying to use the SSH agent.

lapfelix:
Fixes the cloning test through ssh (`GCSingleCommitRepositoryTests`'s `testClone`) on computers that don't have a correctly setup ssh agent, like mine 🥴 (as the keys in `~/.ssh` will be used instead, and that's going to work as long [as they're not protected by a passphrase](https://github.com/git-up/GitUp/issues/35)).

If `SSH_AUTH_SOCK` isn't there, it will fail later on anyway when libgit2 uses libssh2.

Should also make ssh agent workarounds like this one unnecessary: https://github.com/git-up/GitUp/issues/161#issuecomment-643863909

### DIFF
--- a/GitUpKit/Core/GCRepository.m
+++ b/GitUpKit/Core/GCRepository.m
@@ -444,7 +444,7 @@ static int _CredentialsCallback(git_cred** cred, const char* url, const char* us
   GCRepository* repository = (__bridge GCRepository*)payload;
   if (allowed_types & GIT_CREDTYPE_SSH_KEY) {
 #if !TARGET_OS_IPHONE
-    if (!repository->_didTrySSHAgent) {
+    if (!repository->_didTrySSHAgent && getenv("SSH_AUTH_SOCK")) {
       repository->_didTrySSHAgent = YES;
       return git_cred_ssh_key_from_agent(cred, user);
     }


lucasderraugh:
@lapfelix Being relatively unfamiliar with this portion of the codebase, does libssh2 not support accessing passphrased sah keys?

lapfelix:
hmm it would definitely be possible to support passphrased ssh keys, and GitUp could save the passphrase in the keychain. 

(In theory though the ssh agent should "just work" but in practice, I'm not sure why it's not super reliable and why it sometimes needs to be kickstarted after a reboot)

lucasderraugh:
I guess I'm trying to figure out what this particular PR is trying to improve. Wouldn't this environment variable be set as long as ssh-agent was running? Is this just meant to short circuit the case where it isn't?

lapfelix:
Yes basically I didn't find any straightforward way to retry calling the `_CredentialsCallback`, but if the `ssh-agent` isn't running, the environment variable won't be set, so we don't even need to try the `ssh-agent`, we know it'll fail.

This PR helps cases where the ssh agent isn't running, but maybe there's a more "correct" solution to the problem. Like I wonder when line 455 ever executed, for instance

lucasderraugh:
For some reason I seem to be failing in `testClone` with or without this change on the SSH clone `GCSingleCommitRepositoryTests:173` and I'm trying to figure out why but I assume it's related to what you were trying to fix here. I know these tests were all passing previously, not sure if macOS version changed something here but if you know how to get them passing, I'd be happy to hear it.

Cykelero:
It seems that when running tests, the `SSH_AUTH_SOCK` env variable is not defined; whereas when running the app target, the variable does have a value (something like `/private/tmp/com.apple.launchd.A4ChBGpp47/Listeners`). I think that's what preventing libgit2 from cloning the repo.

But, why does that difference exist? I'm not familiar with how Xcode runs things enough, and looking at the schemes, I didn't find anything obvious that'd explain the difference.

lucasderraugh:
Ya it's strange as I know this passed previously without change to this. And yes cloning this repo via SSH in app does work.

lucasderraugh:
I'm going to close this out as I don't see how it helps us out. Seems like ultimately we're not defining the env var properly and we ultimately need a way to resolve that issue.

If anyone feels strongly about the need for this then I'll reopen but doing some cleanup of the PRs otherwise.

PR #988, opened 2024-04-13
Specify `initial_head` when initializing a repository for tests so the global git config doesn't make it fail

lapfelix:
My global git config includes `init.defaultbranch` set to `develop` so two thirds of the tests would fail on my computer. This change makes the tests work on more computers by explicitly specifying "master" as the default branch name (through libgit2's `initial_head` option) so the computer's global git config is ignored.


lucasderraugh (approved):

storagesiddiq:
Thank you for your query

PR #993, opened 2024-04-16
Fix delay when cmd tabbing into GitUp

lapfelix:
So `NSApplicationDidBecomeActiveNotification` isn't blocked so command-tabbing into GitUp isn't delayed because we're refreshing all the repositories synchronously. The work will still be done on the main thread, but it will now be less likely to be noticed (especially with Stage Manager activated, the reload will probably be done by the time the Stage Manager animation is done)

**Steps to reproduce the problem before:**
1. Open several big repositories
2. Switch to the Finder or any other app
3. cmd tab back into GitUp

**Before:**
Takes 1-2 seconds before macOS actually switches to GitUp.

**With the async dispatch:**
macOS instantly switches to GitUp, but the app is unresponsive for 1-2 seconds.

### DIFF
--- a/GitUp/Application/Document.m
+++ b/GitUp/Application/Document.m
@@ -780,12 +780,18 @@ static NSString* _StringFromRepositoryState(GCRepositoryState state) {
 }
 
 - (void)_didBecomeActive:(NSNotification*)notification {
-  [_repository notifyRepositoryChanged];  // Make sure we are up-to-date right now
+  /** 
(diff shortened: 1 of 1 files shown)


Cykelero:
Just wanted to say—that's a really clever fix! A pretty big difference in UX for very little code change :)

drewvolz:
Thank you so much @lapfelix, this fixed the issue for me.

PR #1034, opened 2024-11-23
Null out modification date of index entries when (un)staging lines

Cykelero:
This fixes an issue where GitUp wrote incorrect data to the index, causing both the Git CLI and GitUp itself to produce incorrect diffs.

# Reproducing

- Download the [bear-themes](https://github.com/user-attachments/files/17879934/bear-themes.zip) repo
- The file `Red Graphite.theme` has changes, and is fully staged. In it, unstage the remove/add deltas at line 9.
- The index is now corrupted: GitUp doesn't show those line 9 changes in the workdir nor stage views, and the Git CLI thinks they're still staged.

# Fix

For performance, when producing diffs, Git skips reading the contents of files under some conditions. Here, the bug is that Git skips reading our file because it assumes it's identical between the working directory and the stage. Setting the file's date to null in the index prevents that behavior.

# Notes

This fix has been in use in Retcon since January 2024, with no noticed adverse effect.

### DIFF
--- a/GitUpKit/Core/GCIndex.m
+++ b/GitUpKit/Core/GCIndex.m
@@ -357,7 +357,7 @@ cleanup:
 - (BOOL)addLinesInWorkingDirectoryFile:(NSString*)path toIndex:(GCIndex*)index error:(NSError**)error usingFilter:(GCIndexLineFilter)filter {
   const char* filePath = GCGitPathFromFileSystemPath(path);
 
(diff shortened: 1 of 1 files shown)


lucasderraugh:
Thanks, I'll take a look probably this weekend.

lucasderraugh:
Thanks, looks like it improves that behavior.

PR #1037, opened 2024-12-03
Fix some “Stage All” special cases

Cykelero:
This fixes two issues with the Stage All button.

# Reproducing

Reproducing the conflict issue:
- Cause a conflict from another app (e.g. by running a `git pull`)
- Delete the file that's in conflict
- In GitUp, try to _Stage All_: the operation aborts with a “No such file or directory” error

Reproducing the submodule issue:
- In a repository, copy another repository (just copy a repo folder; don't use submodule features or anything)
- In GitUp, try to _Stage All_ in the outer repository: the operation aborts with a “submodule 'SUB_REPO_PATH' has not been added yet” error

# Notes

The fixes look at the error messages to decide what to do—that might not be ideal, but:
- The specific submodule error is [emitted by libgit2](https://github.com/git-up/libgit2/blob/fc0392bbcdffca9edfb5a09e6d330aa00431d4d3/src/submodule.c#L129), with no distinguishing metadata. To avoid reading the error message, we'd have to essentially reimplement that libgit2 behavior. The fix should hold over time, as libgit2 seems to change existing code fairly rarely, and is English-only. The fallout if the fix does break is minimal.
- The “no such file” error also doesn't have distinguishing features that I could find (its code is a generic `-1`). That one feels flakier, as Apple could change the error message at any time, and the message could be localized if the app runs in a different language; but in practice, Apple is unlikely to change that message, and GitUp is in English only.

### DIFF
--- a/GitUpKit/Extensions/GCRepository+Utilities.m
+++ b/GitUpKit/Extensions/GCRepository+Utilities.m
@@ -205,7 +205,9 @@ NSString* GCNameFromHostingService(GCHostingService service) {
         if (delta.submodule) {
           GCSubmodule* submodule = [self lookupSubmoduleWithName:delta.canonicalPath error:error];
           if (!submodule || ![self addSubmoduleToRepositoryIndex:submodule error:error]) {
-            return NO;
+            if (![[*error localizedDescription] hasSuffix:@"' has not been added yet"]) {
+              return NO;
+            }
           }
         } else {
           if (![self addFileInWorkingDirectory:delta.canonicalPath toIndex:index error:error]) {
--- a/GitUpKit/Extensions/GCRepository+Utilities.m
+++ b/GitUpKit/Extensions/GCRepository+Utilities.m
@@ -209,7 +209,13 @@ NSString* GCNameFromHostingService(GCHostingService service) {
           }
         } else {
           if (![self addFileInWorkingDirectory:delta.canonicalPath toIndex:index error:error]) {
-            return NO;
+            BOOL wasJustTryingToStageADeletedConflictingFile =
+              delta.change == kGCFileDiffChange_Conflicted
+              && [[*error localizedDescription] isEqualToString:@"No such file or directory"];
+            
+            if (!wasJustTryingToStageADeletedConflictingFile) {
+              return NO;
+            }
           }
         }
         break;


GitUpKit/Extensions/GCRepository+Index.m
```diff
@@ -79,14 +79,16 @@ - (BOOL)addFilesToIndex:(NSArray<NSString*>*)paths error:(NSError**)error {
     shouldWriteRepository = YES;
   }
 
-  if (failed && shouldWriteRepository) {
-    if (shouldWriteRepository) {
+  if (shouldWriteRepository) {
+    if (failed) {
```
lucasderraugh, line 83: Hey Nathan (@Cykelero), apologies for taking forever on this. I'm trying to wrap my head around the logic here. Previously if there was any failure we wouldn't write anything, but that's no longer the case. However, the logic in the for loop is to break on a the first failed thing to add.

With the changes made here, is it now more appropriate to `continue` rather than `break` (or else statement)?

lucasderraugh (changes requested):

GitUpKit/Extensions/GCRepository+Index.m
Cykelero: No problem for the delay, of course.

I think you're right about it being a better idea to continue—I had carried over a habit from Retcon (as soon as anything goes wrong, it tries to immediately stop touching anything) but that's not a good fit in GUK. Fixed!

(also, I renamed `shouldWriteRepository` to `needsToWriteIndex`, for clarity)

GitUpKit/Extensions/GCRepository+Utilities.m
```diff
@@ -205,11 +205,19 @@ - (BOOL)syncIndexWithWorkingDirectory:(NSError**)error {
         if (delta.submodule) {
           GCSubmodule* submodule = [self lookupSubmoduleWithName:delta.canonicalPath error:error];
           if (!submodule || ![self addSubmoduleToRepositoryIndex:submodule error:error]) {
-            return NO;
+            if (![[*error localizedDescription] hasSuffix:@"' has not been added yet"]) {
```
lucasderraugh, line 208: Also, is it possible error is NULL here? If I remember my C-isms well enough, dereferencing error would crash, no?

Cykelero: Good catch. I fixed both instances of that bug.

Of note, my fix is a bit too simple: it effectively disables this PR's bugfix, when no error is passed. But, _that_ new bug is niche enough that it should be OK; handling it right would mean more code, and some pointer juggling I'm not completely comfortable with.

lucasderraugh (approved):

Issue #1047, opened 2025-01-22
Very long lines are not drawn correctly.


matrouthier:
When a branch line has many commits, the line is not drawn correctly.  

This can be reproduced on projects with over 6000 commits on the master branch.

<img width="105" alt="Image" src="https://github.com/user-attachments/assets/9edf424b-5067-442c-bfe0-fc6ea73d53a4" />

As the view is scrolled down, we can see that the line is in fact a very long and thin triangle.

<img width="55" alt="Image" src="https://github.com/user-attachments/assets/f45a9a00-015f-478e-b201-01867c370059" />
<img width="52" alt="Image" src="https://github.com/user-attachments/assets/0fedaf1f-b050-4d66-bab7-a29ff3b27b14" />
<img width="52" alt="Image" src="https://github.com/user-attachments/assets/c030d884-f018-4d12-8684-6df65fb54b20" />
<img width="49" alt="Image" src="https://github.com/user-attachments/assets/4f772304-a2c2-4dec-b5d1-146784c2ac9c" />

And the line is corrupted at the bottom too.

<img width="50" alt="Image" src="https://github.com/user-attachments/assets/3bdccc98-fdda-4924-b881-35e996c9749f" />

This looks like the manifestation of a bug in CoreGraphics which occurs when a path is very large and has very small details at the edges.

lucasderraugh:
Closing for #1042

PR #1048, opened 2025-01-22
Each line segment is drawn independently. A fix for issue #1047

matrouthier:


lucasderraugh:
Did this resolve the issue?

matrouthier:
Yes it solves the drawing.  The problem is definitely there.  

However, after a while, when the app is in the background, I get random asserts in metal.  It's weird.  I don't know if it's directly related, but that's why I cancelled the PR.  It would be interesting if somebody else could try it.

Issue #470, opened 2018-07-09
Make font size selectable


jwiegley:
I'd love to use this tool, but everything is too tiny to read on my monitor. Is there a way to increase font size by, say, 100%?

swisspol:
No, it’s not possible and there are no plans to change it.

jwiegley:
Ah, too bad, it looks like it would have been a nice tool to use.

MrLavender:
Yes, this is a real shame. GitUp is exactly what I want from a Git GUI but this one issue makes it unusable for me. The font is seriously tiny by any standard, and while I can read it it's not comfortable for extended use.

Just allowing it to be configured for the diff text in the commit view would make all the difference, since that's what I'm looking at most of the time. Even the standard git gui client allows this.

donraj:
Can we have the option to change font size please? On big monitors its not really usable.

douglashill:
I've implemented a text size setting in my fork: https://github.com/douglashill/GitUp/releases

comatory:
Wish at least the diff view would be configurable.

mluisbrown:
I created a PR for this: https://github.com/git-up/GitUp/pull/2727

PR #2727, opened 2025-10-23
Make the Text Size preference apply to the text in the diff view, not just the line numbers

mluisbrown:
The Text Size setting in the preference pane currently only changes the size of the line numbers in the diff viewer, which is not that useful. Presumably it is supposed to also change the text size in the diff content.

This PR makes that change. The Text Size setting will now correctly change the size of the text in the diff content, not just the line numbers. 

This addresses https://github.com/git-up/GitUp/issues/470

Below are two screen shots with this change using the smallest and largest text settings.

GitUp2
GitUp1

### DIFF
--- a/GitUpKit/Interface/GISplitDiffView.m
+++ b/GitUpKit/Interface/GISplitDiffView.m
@@ -108,6 +108,7 @@ typedef NS_ENUM(NSUInteger, SelectionMode) {
 @implementation GISplitDiffView {
   NSMutableArray* _lines;
   NSSize _size;
+  CGFloat _layoutFontSize;
 
   BOOL _rightSelection;
   NSMutableIndexSet* _selectedLines;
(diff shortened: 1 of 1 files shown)


lucasderraugh (approved):
Good catch! This seems like a reasonable approach without having us unnecessarily calling logic in `-updateLayoutForWidth:`

lucasderraugh:
To be clear, this was only a bug when updating while the view was visible, it would work if you updated outside of this being presented. Looks like UnifiedDiffView doesn't have the same checks so it already dynamically changes.

mluisbrown:
> To be clear, this was only a bug when updating while the view was visible, it would work if you updated outside of this being presented. 

I tried changing the setting without the diff view being visible, but it still only changed the line number sizing. 

However, not important now.

PR #2771, opened 2026-04-23
Fix crash when staging lines with the index locked

### LINKED ISSUE #2768: Crash in GCDiff _cacheDeltasIfNeeded when git index is locked (GIT_ELOCKED)
(via claude)

## Crash Report

GitUp crashes with `NSInvalidArgumentException: capacity (18446744073709551615) is ridiculous` in `-[GCDiff _cacheDeltasIfNeeded]` when another process holds the git index lock.

### Exception

```
Exception Type:    EXC_CRASH (SIGABRT)
Exception Reason:  *** -[__NSPlaceholderArray initWithCapacity:]: capacity (18446744073709551615) is ridiculous
```

### Crash Backtrace

```
0  CoreFoundation  -[__NSPlaceholderArray initWithCapacity:] + 376
1  GitUpKit        -[GCDiff _cacheDeltasIfNeeded] + 92
2  GitUpKit        -[GCDiff deltas] + 20
3  GitUpKit        -[GIAdvancedCommitViewController _reloadContents] + 132
4  GitUpKit        -[GIAdvancedCommitViewController repositoryStatusDidUpdate] + 68
5  ...
6  GitUpKit        -[GCLiveRepository _updateStatus:] + 1068
7  GitUpKit        -[GCLiveRepository _notifyWorkingDirectoryChanged:gitDirectoryChanged:] + 344
```

### Root Cause Analysis

The crash is caused by a chain of events when the git index file is locked by another process:

1. `git_diff_index_to_workdir()` is called with the `GIT_DIFF_UPDATE_INDEX` flag (in `GCDiff.m`, both `diffWorkingDirectoryWithIndex:` and `diffWorkingDirectoryWithCommit:usingIndex:`)
2. The diff computation succeeds, but `git_index_write()` fails with `GIT_ELOCKED` because another process holds the lock
3. In the bundled libgit2 (`diff_generate.c:1513-1518`), on `GIT_ELOCKED` after diff computation, the function jumps to its cleanup label which **frees the computed diff** and leaves `*out = NULL`
4. GitUp's code overrides `GIT_ELOCKED` → `GIT_OK`, assuming the diff pointer is still valid — but it is now `NULL`
5. A `GCDiff` object is created wrapping a **NULL** `git_diff*` pointer
6. Later, `git_diff_num_deltas(NULL)` triggers `GIT_ASSERT_ARG(diff)` in `assert_safe.h`. In release builds, this macro **returns -1** rather than crashing
7. Since `git_diff_num_deltas` returns `size_t` (unsigned), `-1` becomes `18446744073709551615` (`0xFFFFFFFFFFFFFFFF`)
8. `[[NSMutableArray alloc] initWithCapacity:0xFFFFFFFFFFFFFFFF]` throws the "capacity is ridiculous" exception

### Affected Code

**`GCDiff.m` — `diffWorkingDirectoryWithIndex:` (line ~600):**
```objc
int status = git_diff_index_to_workdir(outDiff, self.private, index.private, diffOptions);
if (status == GIT_ELOCKED) {
    status = GIT_OK;  // BUG: *outDiff is NULL here, diff was freed by libgit2
}
```

**`GCDiff.m` — `diffWorkingDirectoryWithCommit:usingIndex:` (line ~562):**
Same pattern — also has a secondary bug where `git_diff_merge(*outDiff, diff2)` is called with a NULL `diff2`, then `*outDiff` is freed but the pointer is not NULLed out, causing a double-free in `_diffWithType:`'s cleanup.

### Proposed Fix

**Primary fix:** When `git_diff_index_to_workdir` returns `GIT_ELOCKED`, retry without the `GIT_DIFF_UPDATE_INDEX` flag to obtain a valid diff:

```objc
int status = git_diff_index_to_workdir(outDiff, self.private, index.private, diffOptions);
if (status == GIT_ELOCKED) {
    // On GIT_ELOCKED, libgit2 frees the diff and sets *outDiff to NULL.
    // Retry without the flag to get a valid diff.
    diffOptions->flags &= ~GIT_DIFF_UPDATE_INDEX;
    status = git_diff_index_to_workdir(outDiff, self.private, index.private, diffOptions);
}
```

**Secondary fix (defense-in-depth):** Add a NULL check in `_cacheDeltasIfNeeded`:

```objc
- (void)_cacheDeltasIfNeeded {
  if (_deltas == nil) {
    if (_private == NULL) {
      XLOG_DEBUG_UNREACHABLE();
      _deltas = [[NSMutableArray alloc] init];
      return;
    }
    size_t count = git_diff_num_deltas(_private);
    // ...
```

**Tertiary fix:** In the `diffWorkingDirectoryWithCommit:` block, NULL out `*outDiff` after freeing it to prevent double-free:

```objc
if (status != GIT_OK) {
    git_diff_free(*outDiff);
    *outDiff = NULL;  // Prevent double-free in _diffWithType: cleanup
}
```

### Environment

- GitUp 1.4.3 (build 1052)
- macOS 26.4.1 (25E253), ARM64

Cykelero:
Fixes #2768.

# Why it crashes

See the linked issue for a LLM-generated attempt at diagnosing the issue; it seems rather correct. (the proposed solutions are pretty terrible though)

In a nutshell: in two spots*, GUK calls `git_diff_index_to_workdir()`, and _explicitly ignores its returned error_ if it’s `GIT_ELOCKED`. The reasoning is that we don’t care about failing to write the index: that’s just an optional, beneficial side-effect to our actual need, which is to get a diff.

However, when failing in that way, `git_diff_index_to_workdir()` does _not_ return a diff**, it returns null! And so that null value is returned, silently makes its way through GUK, until this causes an array to be allocated with an incorrect size.

*`diffWorkingDirectoryWithCommit` and `diffWorkingDirectoryWithIndex`, on `GCRepository`.
** Presumably it did at some point, explaining why the code is written that way.

# Reproducing

To reproduce, I had to patch `git_diff_index_to_workdir()` to always write the index:
```diff
-	if ((diff->opts.flags & GIT_DIFF_UPDATE_INDEX) && ((git_diff_generated *)diff)->index_updated)
+	if ((diff->opts.flags & GIT_DIFF_UPDATE_INDEX))
```

Then, you can trigger the crash by staging an individual line. (that’s what I did in Retcon, anyway, but it should work in GitUp as well)

# The fix

I removed the override altogether. When I tested, if `git_diff_index_to_workdir()` returns `GIT_ELOCKED`, it always returns an empty diff. Maybe that’s not always the case, but it doesn’t feel like we should trust its return value in these cases anyway.

### DIFF
--- a/GitUpKit/Core/GCDiff.m
+++ b/GitUpKit/Core/GCDiff.m
@@ -560,9 +560,6 @@ cleanup:
                                    git_diff* diff2;
                                    diffOptions->flags |= GIT_DIFF_UPDATE_INDEX;
                                    status = git_diff_index_to_workdir(&diff2, self.private, index.private, diffOptions);
-                                   if (status == GIT_ELOCKED) {
-                                     status = GIT_OK;  // Passing GIT_DIFF_UPDATE_INDEX means git_diff_index_to_workdir() may attempt to write the index and this could fail if it is currently locked by another process but that's OK to ignore this failure
-                                   }
                                    if (status == GIT_OK) {
                                      status = git_diff_merge(*outDiff, diff2);
                                      if (status != GIT_OK) {
@@ -599,11 +596,7 @@ cleanup:
                        error:error
                        block:^int(git_diff** outDiff, git_diff_options* diffOptions) {
                          diffOptions->flags |= GIT_DIFF_UPDATE_INDEX;
-                         int status = git_diff_index_to_workdir(outDiff, self.private, index.private, diffOptions);
-                         if (status == GIT_ELOCKED) {
-                           status = GIT_OK;  // Passing GIT_DIFF_UPDATE_INDEX means git_diff_index_to_workdir() will attempt to write the index even if there are no changes and this could fail if it is currently locked by another process
-                         }
-                         return status;
+                         return git_diff_index_to_workdir(outDiff, self.private, index.private, diffOptions);
                        }];
 }
 


lucasderraugh:
@Cykelero What happens when this path is hit and GIT_ELOCKED is returned? Do we throw some kind of error UI?

Cykelero:
Yeah, the usual dialog:
image

lucasderraugh:
Thanks!

PR #2775, opened 2026-05-11
Fix virtual graph lines claiming shared history

### LINKED ISSUE #336: Visual glitch in graph, thin dashed line all the way
I know this feels a bit _anal retentive_, please bear with me…

It seems that if the last changed branch is not a tip, the whole line graph for that branch is thin and dashed.

screen shot 2017-06-28 at 16 49 18

In the case of this screenshot, I rebased `pixart` on top of `master`, which created a bold branch to the top left (before it was to the right, being a branch spun out of master earlier on, the remote-tracking branch is still there).

I also had a branch called `import-new-features` which was before the tip of `pixart` (also seen on the right with the remote-tracking branch). To update it's location, I used the "Set Tip" on the corresponding commit of the newly rebased branch. This cause the branch `import-new-features` to be the left-most, and since it's not a bold-line change, it has a thin dash line. The problem is that it doesn't stop at the juncture with `pixart`, it goes all the way down to the beginning of time, and you can see white breaks due to the commits dots. I would need to make some commits on either branch to solve the visual glitch.

Another way to go about it would be rebasing `import-new-features` instead of `pixart`, set `pixart`'s tip to `import-new-features` and then cherry-pick the commits from the remote-tracking branch. With a couple of commits it's fine, with a more a complex history it becomes a bit tedious.

Shriram-Vasudevan:
Fixes #336.

## Summary
- process real graph nodes before dummy/virtual branch-tip nodes when generating each layer
- prevents a virtual branch tip from becoming the primary line for a shared commit and rendering the rest of its ancestry as thin/dashed
- add a regression test with out-of-order committer dates, matching the rebase/amend scenario described in the issue

## Verification
- `clang -fsyntax-only ... GitUpKit/Interface/GIGraph.m`
- `git diff --no-index --check GitUp-baseline GitUp`

Could not run the XCTest suite locally because this machine only has Command Line Tools installed; `xcodebuild` requires full Xcode. `clang-format` is also unavailable locally.

### DIFF
--- a/GitUpKit/Interface/GIGraph-Tests.m
+++ b/GitUpKit/Interface/GIGraph-Tests.m
@@ -23,10 +23,61 @@
 #define kNotationSeparator @"##### NOTATION #####\n\n"
 #define kGraphSeparator @"##### GRAPH #####\n\n"
 
-@interface GIGraphTests : XCTestCase
+@interface GIGraphTests : GCTestCase
 @end
 
 @implementation GIGraphTests
+
+- (NSString*)_runGitWithRepository:(GCRepository*)repository arguments:(NSArray*)arguments environment:(NSDictionary*)environment {
+  GCTask* task = [[GCTask alloc] initWithExecutablePath:@"/usr/bin/git"];
+  task.currentDirectoryPath = repository.workingDirectoryPath;
+  task.additionalEnvironment = environment;
+  NSData* data;
+  BOOL success = [task runWithArguments:arguments stdin:nil stdout:&data stderr:NULL exitStatus:NULL error:NULL];
+  XCTAssertTrue(success);
+  return [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
+}
+
+- (void)_makeEmptyCommitInRepository:(GCRepository*)repository message:(NSString*)message date:(NSString*)date {
+  NSDictionary* environment = @{@"GIT_AUTHOR_DATE" : date, @"GIT_COMMITTER_DATE" : date};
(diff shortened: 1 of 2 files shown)


lucasderraugh:
Thanks for your submission. However, this appears to introduce new issues. Please verify improvements and include screenshots for visual changes like this before submitting PRs.

I'll leave this open for some time, but after a week I will likely close it if relevant fixes aren't made.

Before is left, this PR is right.
Screenshot 2026-05-11 at 4 34 59 PM

Shriram-Vasudevan:
Pushed follow-up commit 6ed58b7 to address this regression. The original fix was too broad because it processed every non-dummy node before every dummy node, which changed normal graph lane ordering such as the existing simple fixture. The new version only postpones lines whose primary line is actually virtual, so ordinary dummy connector ordering stays unchanged while virtual branch tips still cannot claim shared ancestry as their primary dashed line.

Local verification:
- `clang -fsyntax-only -fobjc-arc -fmodules -DDEBUG=1 -D__GI_HAS_APPKIT__=0 ... GitUpKit/Interface/GIGraph.m` -> passed
- `git diff --check` -> passed

I still cannot run the full `xcodebuild test` suite or produce a local app screenshot on this machine because it only has Command Line Tools installed, not full Xcode/XCTest. The pushed commit should let the GitHub macOS CI verify the existing graph fixture plus the new regression test.

lucasderraugh:
If you can't run and test this locally, I don't have the patience to debug for you. Frankly, this account smells of AI and I will not hesitate to ban you from this repo if there are further followups in AI.

PR #2786, opened 2026-06-10
Move back to CMake for libgit2 compilation

### LINKED ISSUE #2773: Zombie processes on 1.5.0 beta
Recently I've received errors when opening new Terminals:
```
[forkpty: Resource temporarily unavailable]
Could not create a new process and open a pseudo-tty.
```

While investigate if this was caused by exceeding the max number of processes I noticed that there were a *lot* of zombie processes (thousands):

```
$ ps -axo pid,ppid,state,command | grep defunct
114  1436 Z    <defunct>
115  1436 Z    <defunct>
156  1436 Z    <defunct>
194  1436 Z    <defunct>
....
99943  1436 Z    <defunct>
99946  1436 Z    <defunct>
99998  1436 Z    <defunct>
```

Which were caused by GitUp:

```
$ ps -p 1436 -o pid,ppid,command
PID  PPID COMMAND
1436     1 /Applications/GitUp.app/Contents/MacOS/GitUp
```

This happens using 1.5.0 beta on macOS 15 Sequoia. I can reproduce it by launching GitUp and opening some specific repository, this causes a zombie process to show up. Closing the window and opening the same repo gives another zombie process, etc. Some repos are not affected, I'm not sure what exactly triggers this. Will try to do some more tests to see if I can narrow it down.

Reverting to 1.4.3 stable release seems to fix the issue.

lucasderraugh:
At some point we tried to re-engineer the build process of libgit2 but this comes with lots of sharp edges with the only benefit being automatic building of dependencies through SPM.

Ultimately, I don't think this is worth it given the complexity of managing the various parameters for compilation. This PR reaches a middle ground where we will pre-compile the libgit2 project using cmake into xcframework.

Also fixes #2773 by using a flag we used previously -DUSE_SSH=libssh2. Ultimately this will use libssh2 rather than the external ssh option (ssh_exec). Looking into it, there was a bug in the libgit2 implementation that would never wait on the process on close. I think there is a patch that could be made but I believe this change will return to the behavior of 1.4.3.


Cykelero:
Reverting to libssh2 comes at a heavy cost! We’ll lose all support for custom `.ssh/config` files, which includes easy third-party agent integration (e.g. 1Password). This was [a big reason](https://retcon.app/releases#retcon-1.3.0) for [updating our libgit2 fork](https://github.com/git-up/GitUp/pull/2683) last year.

I assume some GitUp users will miss that support, and for me personally, it also means I’ll need to manage different libgit2 build settings in Retcon, since dropping SSH config support is not an option anymore.

Do you think there might be another way of fixing the zombie issue? I don’t think the tradeoff of switching back to libssh2 is worth it.

lucasderraugh:
Ah, thank you. I was trying to recall why we would have moved away. I had a patch in that I think worked so let me retry using that.

PR #2803, opened 2026-07-23
Leave working dir alone when unable to check out branch checked out in worktree

### LINKED ISSUE #2713: GitUp leaves behind many incorrectly modified files when checkout of HEAD branch from alternate worktree fails (as expected)
I have a project where I have 2 worktrees on the same repo.
I often forget which branch I have checked out in the alternate worktree.
So when I use GitUp to switch branches, sometimes I'll receive the error:

Image

This is fine and expected (if a little more cryptic than the git CLI would print) `fatal: '<branch>' is already checked out at '<location>'`

However, unfortunately when this happens GitUp also leaves a bunch of erroneously modified files left behind that have to be cleaned up.

Image

These staged changes contain what would be a "revert" against the other branch's HEAD.

A similar issue happens sometimes when a merge fails, or a rebase fails, but its probably the same piece of code.

This behavior does not happen when an attempted checkout fails from the git cli. It just prints the quoted error message and no code is left in a weird state.

rodovich:
### Summary

Fixes #2713.

Previously, `checkoutLocalBranch:` wrote the target tree to the working directory and index before moving HEAD. When the branch is checked out in a linked worktree, `git_repository_set_head()` fails, leaving the applied tree changes behind as staged debris.

### Testing

Added a test that checks out a branch in a worktree and then calls `checkoutLocalBranch:` and validates that the working directory is unchanged. Also tested manually.

GitUp 1.5.0 (1058): checking out the branch fails but leaves modified files in the index:

image
image

Built from this branch: checking out fails (with a slightly clearer message) and does not touch the index:

image
image

----

### DIFF
--- a/GitUpKit/Core/GCRepository+HEAD-Tests.m
+++ b/GitUpKit/Core/GCRepository+HEAD-Tests.m
@@ -194,4 +194,37 @@
   [self assertContentsOfFileAtPath:@"new_file.txt" equalsString:@"New Content\n"];
 }
 
+// Regression test for issue #2713: checking out a branch that is already checked out in another
+// worktree must fail *without* mutating the working directory or index. Previously the target tree
+// was written to disk before the (failing) HEAD move, leaving staged "revert" debris behind.
+- (void)testCheckoutBranchCheckedOutInLinkedWorktree {
+  // We start on master with a clean working directory.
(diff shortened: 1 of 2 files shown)


lucasderraugh:
Thanks for the contribution. With worktrees rising in popularity thanks to AI tooling, I'm sure we have much to do to catch up.

I'm out this week but I'll working through my backlog of PRs next week.

lucasderraugh (approved):

PR #2812, opened 2026-08-11
Fix delete key deleting commit when clearing search field

### LINKED ISSUE #991: Search field delete/backspace should not carry through to deleting commits
When I'm typing in the search field, and want to delete everything I've typed there and start a new search, I usually am lazy and press and hold down delete key. 
Currently, when the search field runs out of characters it automatically closes and transfers focus back to the main GitUp window. Which, if I'm holding down delete still, will then attempt to delete the HEAD commit of my current branch. Thankfully there is a confirmation box for that, but I would wager the `delete commit` action is NEVER what any dev would want to happen in that scenario.

Essentially this comes down to a focus issue, the search box probably should not instantly & automatically close thereby transferring active context back to the main window when the user is still using the keyboard.
If the user clicks the `X` button within the search field that makes sense to completely clear and close out the search sidebar, but not any sort of keyboard entry. My cursor should remain within the search field unless I move out of it (`tab` key or a click of some kind)

YuriNachos:
Fixes #991

## Summary

When the map view search field is emptied by editing (for example, holding the Delete/Backspace key to clear what was typed), the still-pressed Delete keystroke is no longer forwarded to the map view, so it can no longer accidentally delete the selected (HEAD) commit. Explicit dismissal of the search (Escape key or the search field cancel button) still returns focus to the map view as before.

## Root Cause

`-performSearch:` is sent on every edit of the toolbar search field. Its empty-query branch unconditionally called `makeFirstResponder:` to move key focus to the map view. As soon as the last character was deleted, focus left the field while the Delete key was still held down, so the next repeated Delete event was delivered to the map view, which binds Delete to "delete the selected commit" (issue #991).

## Changes

- `GitUp/Application/Document.m`
  - In `-performSearch:`, only transfer focus to the map view when the search field is not currently being edited. When the query is cleared by editing, focus stays in the (now empty) search field, so a held Delete key cannot reach commit deletion.
  - In `-closeSearch:`, explicitly return focus to the map view so Escape / the cancel button keep their existing dismiss-to-map behavior.

## Test plan

- `xcodebuild build -scheme "Application" -configuration "Release" -derivedDataPath /tmp/gitup-dd CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO` succeeds (compile-checked locally).
- Manual repro from the issue: in a repository, focus the toolbar search field, type a query, then press and hold Delete until the field is empty. Before the change the held Delete deletes the HEAD commit (after a confirmation); after the change focus remains in the search field and no commit deletion is triggered. Pressing Escape still dismisses the search and returns focus to the map.

This is a focus-handling change in the AppKit layer, so it is not covered by the headless `GitUpKit (macOS)` unit tests; the full build and CI suite validate compilation.


lucasderraugh:
Sorry for taking this a bit, but there was slightly different behavior I wanted for resolving this. Changes was made in 84704ac961427544608483ef0b71f0bdb92d9dc3.

PR #2813, opened 2026-08-11
Resolve repository hooks from the common directory

### LINKED ISSUE #1053: GitUp does not execute pre-commit/pre-push hook for worktrees
The title more or less says it all.
When I have a worktree open in gitup, and I commit or push, it always succeeds.
But the same changeset when I have the base repo open in gitup, fails as expected.
As a note: in terminal both trigger the hooks as expected.

Can easily be tested with the following pre-commit hook.
```
#!/bin/bash
echo 'test pre-commit'
exit 1
```
Same issue with pre-push.

I could have sworn this worked at some point, but maybe I'm mis-remembering.

For reference 
`git config --local --list`:
```
core.repositoryformatversion=0
core.filemode=true
core.bare=false
core.logallrefupdates=true
core.ignorecase=true
core.precomposeunicode=true
extensions.worktreeconfig=true
```
and then in the worktree
`git config --worktree --list`
```
core.hookspath=~/my_project/.git/hooks
```

YuriNachos:
Fixes #1053

## Summary

`pre-commit`, `pre-push`, and other Git hooks are now found when the repository is a linked worktree. Previously GitUp silently skipped hooks for worktrees because it looked them up in the wrong directory.

## Root Cause

`-pathForHookWithName:` built the default hooks directory from `self.repositoryPath`, i.e. `git_repository_path()`. For a linked worktree that returns the per-worktree git directory (e.g. `<main>/.git/worktrees/<name>`), which never contains a `hooks` directory. Git stores hooks in the repository **common** directory (`<main>/.git/hooks`), shared by all linked worktrees, so GitUp never found them and `runHookWithName:` (used for the pre-push hook in `-sendLocalBranchesToRemote:...`) silently treated them as absent.

## Changes

- `GitUpKit/Core/GCRepository.m`
  - In `-pathForHookWithName:`, derive the default hooks path from `git_repository_commondir(_private)` instead of `self.repositoryPath`. For a plain repository the common directory equals the git directory, so behavior is unchanged; only linked worktrees are affected.

## Test plan

- Added `testPathForHookWithName_FoundFromLinkedWorktree` to `GCRepository-Tests.m` (`GCMultipleCommitsRepositoryTests`): creates a linked worktree via `git worktree add`, installs an executable `pre-commit` hook in the main repository hooks directory, and asserts the worktree resolves the hook to that shared path.
- Verified red→green locally:
  ```
  cd GitUpKit
  xcodebuild test -scheme "GitUpKit (macOS)" -destination "platform=macOS" \
    -only-testing:Tests/GCMultipleCommitsRepositoryTests/testPathForHookWithName_FoundFromLinkedWorktree \
    CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO
  ```
  The new test fails on `master` (worktree cannot resolve the hook) and passes with this change.
- Existing `testPathForHookWithName`, `testPathForHookWithName_AbsoluteCustomHooksPath`, and `testPathForHookWithName_RelativeCustomHooksPath` are unaffected (a plain repo has `commondir == gitdir`, and the custom `core.hooksPath` branch is untouched).


SamuelLenyakopiro (approved):

lucasderraugh:
Looks good and straight forward.

